PR #16, opened 2021-12-12
Remove base.wgsl

pixelspark:
Seems as if it is unused


haixuanTao:
Hey pixelspark! Thanks for the contribution! I think it is ok to merge.

I will try to make this package cleaner from this point onward as I will no longer be the sole user.

I had searched for a long time to move tera to compile time but did not now how as it is eating about 120ms.

I will merge my branch `feature/optimisation_graph` with a deep refactoring after yours.

pixelspark:
> Hey pixelspark! Thanks for the contribution! I think it is ok to merge. Let me know if you have any commit you want to add or I'll merge this version.
> 
> I will try to make this package cleaner from this point onward as I will no longer be the sole user.
> 
> I had searched for a long time to move tera to compile time but did not now how as it is eating about 120ms.
> 
> I will merge my branch `feature/optimisation_graph` with a deep refactoring after yours.

Hi @haixuanTao, these PRs were just a few changes I made while getting acquainted with the code. I wanted to write an ONNX runtime on wgpu but you beat me to it :-) If I can find the time I might contribute some more later on.  Feel free to merge in any way you like the best.

About tera compilation: my change simply copies the template files to string literals (hence they end up in the binary), but compilation still happens at runtime (each time you load the templates first due to the `lazy_static!`). I don't think tera provides a way to compile the templates at compile time - if there is a way to serialize tera templates to some efficient binary formats then you could use `build.rs` but I don't think tera supports serialization of the parsed template either...

haixuanTao:
I see! Yeah, I feel like tera is not the most performant templating engine. But the API is very rich and as everything happen outside of the run I'm fine with the trade off.

Ahah, yeah, well I was working on the C++ wrapper of the Microsoft version `onnxruntime-rs` but someone also beat me to it, ahah. I asked him to be a maintainer but he let the question sleep so I had to find something else :) 

The good news is that there is tons of package still left to build in the Rust community even from unmaintained peers. 

Thanks for the contribution! It really means a lot to me as this is my first OSS Package :)

haixuanTao:
So from my benchmark, the user of lazy_static does speed up a bit the loading process. I guess it might be linked to less IO. 

Thanks !

PR #21, opened 2021-12-29
Fix/tidying up

pixelspark:


haixuanTao:
Great! Thanks for cleaning up!

PR #25, opened 2021-12-30
feat: implement a very slow softmax

pixelspark:
Currently the Softmax operator just copies data from input to output. To match Keras, I implemented actual Softmax in WGSL. 

This is the first time I have written any WGSL and my suspicion is that this is probably the slowest possible way to implement it (i.e. single thread, fully serial, not running anything in parallel on the GPU at all - main CPU is probably faster at this!). Nevertheless I think it is good to have this at this point (can make it fast later). 

Note, the actual ONNX spec may also require Softmax on a single axis in a matrix, see [here](https://github.com/onnx/onnx/blob/master/onnx/backend/test/case/node/softmax.py). Probably also should write some tests.

### DIFF
--- a/src/compiler.rs
+++ b/src/compiler.rs
@@ -101,12 +101,13 @@ pub fn compile(
             1,
         ),
         // Copy data
-        "Reshape" | "Dropout" | "Flatten" | "Squeeze" | "Softmax" => (
+        "Reshape" | "Dropout" | "Flatten" | "Squeeze" => (
             "endomorphism/copy.wgsl".to_string(),
             ceil(output_lengths[0], 16) as _,
             1,
             1,
         ),
+        "Softmax" => ("endomorphism/softmax.wgsl".to_string(), 1, 1, 1),
         // Arithmetic operation
         "Add" | "And" | "Div" | "Equal" | "Greater" | "GreaterOrEqual" | "Less" | "LessOrEqual"
         | "Mod" | "Mul" | "Or" | "Sub" => {
--- a/src/compiler.rs
+++ b/src/compiler.rs
@@ -18,6 +18,7 @@ pub fn compile(
     node: &crate::onnx::NodeProto,
     dims_infos: &HashMap<String, Vec<i64>>,
     tera: &Tera,
+    opset_version: i64,
 ) -> CompiledNode {
     // Escape unwanted characters
     let mut inputs = node.get_input().to_vec();
@@ -90,6 +91,7 @@ pub fn compile(
     context.insert("i_chunks", &input_chunks);
     context.insert("o_chunks", &output_chunks);
     context.insert("op_type", &node.get_op_type());
+    context.insert("opset_version", &opset_version);
 
     let (template, x, y, z) = match node.get_op_type() {
         // Map simple function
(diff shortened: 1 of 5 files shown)


src/compiler.rs
```diff
@@ -97,12 +97,13 @@ pub fn compile(
...
         ),
         // Copy data
-        "Reshape" | "Dropout" | "Flatten" | "Squeeze" | "Softmax" => (
+        "Reshape" | "Dropout" | "Flatten" | "Squeeze" => (
             "endomorphism/copy.wgsl".to_string(),
             ceil(output_lengths[0], 16) as _,
             1,
             1,
         ),
+        "Softmax" => ("endomorphism/softmax.wgsl".to_string(), 1, 1, 1),
```
haixuanTao, line 106: So the `1, 1, 1` correspond to the number of threads you want to spawn. By doing this your only spawning 1 threads. I think the number of thread should be `output_lengths[0], 1, 1` although this will  only work if the axis of softmax is the first one.

pixelspark: Hm, not sure I fully understand this yet. When 1 thread is spawned, I assume this means that only one instance of my shader will be running at any time (never in parallel). Then I guess this also means the shader will be invoked multiple times in series (as otherwise my current example would probably not work at all as only the first few elements would be calculated?).

It does seem a bit wasteful that each shader invocation is calculating max and sum each time but I guess the only way to fix that is to have multiple passes (which is more complex). 

Re axis, I think this is configurable for softmax, would have to add a check perhaps.

haixuanTao: Yep, so within the wgsl shader, you have the global_id index correspond to the thread of the shader. Each thread is independent of one another. If the threads are `1, 1, 1` only one thread is going to spawn and the only global_id that is going to spawn is the global_id = (0, 0, 0). If you want to compute the value for global_id from 0 to `output_lengths[0]` you have to specify it in advance. Therefore the `output_lengths[0], 1, 1`. 

I agree that it is quite wasteful, to do the summation in every thread. But, I'm ok with the current status. The thing is that as long as the softmax does not correspond to a large portion of the computation, like convolution for example. Doing optimisation will not result in visible gain.

pixelspark: Thanks for the explanation, very helpful! 👍 

Now I'm just wondering why the current version even works at all. In my toy MNIST model there is a softmax outputting (10,), that should not be possible unless somehow the shader considers 10 elements at once (?) 🤔

pixelspark: > Now I'm just wondering why the current version even works at all. In my toy MNIST model there is a softmax outputting (10,), that should not be possible unless somehow the shader considers 10 elements at once (?) 🤔

Never mind, I am simply looping from 0 to `i_lens[0]` of course, setting every element. 

So with multiple threads you would change the last loop to only set a few of the elements right? (Then i'm not sure having multiple threads will be faster because the only thing that is done in parallel is this last step, where as the sum/max are all the same because they consider all elements). It would be faster for cases with multiple axes however (and each axis getting their own thread).

haixuanTao: ah yes! Missed that part. Ok, well let's leave it as is.

haixuanTao:
Can you just add the attribute axis, and check if the axis is the axis 0. And default the axis attribute to 0 if missing.

To avoid having unknown error.

Thanks in advance.

pixelspark:
> Can you just add the attribute axis, and assert if the axis is the axis 0. And default the axis attribute to 0 if missing.
> 
> To avoid having unknown error.
> 
> Thanks in advance.

Good idea, will do so later today/tomorrow.

pixelspark:
@haixuanTao I added a check on the `axis` attribute. Turns out depending on the ONNX spec version the attribute accepts different values. Therefore I also added a way to retrieve the ONNX opset version referenced, this will be useful for other ops as well.

PR #27, opened 2022-01-05
Fix/binding indexes

pixelspark:
This PR clarifies the 'optimisation' code a bit and also fixes an issue where incorrect binding indexes are applied to inputs when there are more than 4 (for outputs the binding index was defined as `binding_counter % 4` but it was just `binding_counter` for inputs).

### DIFF
--- a/src/optimisation.rs
+++ b/src/optimisation.rs
@@ -189,9 +189,11 @@ pub fn load(
         let mut entries = vec![];
 
         for tensor in current_node.get_input() {
+            // Bindings are numbered 0...3 (MAX_BINDINGS_PER_GROUP-1) in binding groups (starting at group 0)
+            let binding_index = (binding_counter % MAX_BINDINGS_PER_GROUP) as u32;
             entries.push(wgpu::BindGroupEntry {
-                binding: binding_counter,
                 resource: inner_infos
+                binding: binding_index,
                     .get(tensor.as_str())
                     .unwrap_or_else(|| {
                         panic!("Tensor {} is not present in the inner infos", tensor)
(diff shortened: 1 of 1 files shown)


src/optimisation.rs
```diff
@@ -112,17 +112,19 @@ lazy_static! {
     };
 }
 
+const MAX_BINDINGS_PER_GROUP: usize = 4;
```
haixuanTao, line 115: Can we use: https://docs.rs/wgpu/latest/wgpu/struct.Limits.html wgpu limits constant with the defaults level, using `lazy_static`?

pixelspark: That seems like a good idea. However we should probably get the actual limits from the adapter/device at runtime (see [here](https://docs.rs/wgpu/latest/wgpu/struct.Adapter.html#method.limits)).

pixelspark: On second thought we should probably just stick to defining our own limit as 4: our shaders are all written with the maximum of 4 in mind, so even if wgpu would increase the limits we cannot simply assume it without changing the shader source (i.e. group/binding id's would have to be changed).

We could of course make this dynamic by getting the limit at runtime, adding that limit as a template variable, and dynamically determining group/binding id's. Unless there is a great performance benefit in having larger or less groups I am not sure that is worth the trouble.

haixuanTao: Well, the benefit of large bind group is not very clear from my point of you at the moment.

Well, agreed that the shaders are now written for bind group of size 4 and making it dynamic might break things. We can leave it as is for now, and change it later if we want to do optimisation on bind groups.

I can merge then, if it is ok for you.

Thanks for the good discussion.

pixelspark:
OK, should be ready to merge now without conflicts! 👍

PR #31, opened 2022-01-09
feat: support Unsqueeze and Identity ops

pixelspark:
These ops appear to only change shape (which we don't care about as we require the dimensions to be inferred up front anyway), not data. Squeeze was already there so I added Unsqueeze and Identity as well.

We should probably not bother copying data between buffers but have a step in the compiler that simply aliases buffers, i.e.  in `input_a -> Identity -> SomeOtherOp`, the `SomeOtherOp` should just take `input_a` as input rather than the output from identity (which we then won't generate).

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,7 +204,7 @@ fn test_matmul_square_matrix() {
 |<a href="#GridSample">GridSample</a>|<a href="Changelog.md#GridSample-16">16</a>|
 |<a href="#HardSigmoid">HardSigmoid</a>|<a href="Changelog.md#HardSigmoid-6">6</a>, <a href="Changelog.md#HardSigmoid-1">1</a>|
 |<a href="#Hardmax">Hardmax</a>|<a href="Changelog.md#Hardmax-13">13</a>, <a href="Changelog.md#Hardmax-11">11</a>, <a href="Changelog.md#Hardmax-1">1</a>|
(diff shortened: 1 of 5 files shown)


pixelspark:
This PR now includes logic to skip identity ops altogether (when they are followed by some other op) by stringing together the input of the identity op and the input of the next op. The copy shader is still executed when an identity op is last, or is the only op in the graph (tested by the additional test).

haixuanTao:
Would it possible to just skip the node. The thing is by passing on the next node. There is a chance that we miss on some sequence optimisiation.

pixelspark:
> Would it be possible to just skip the node?. The thing is by passing on the next node. There is a chance that we miss on some sequence optimisiation.

Technically the current implementation already skips the node, only it does so inside the sequencer. Are you suggesting to instead add a preprocessing step before we feed the model into the sequencer, which removes these nodes (by editing the ModelProto basically)? Fine with me (just a little bit more fiddly I guess), I can have a go at this later hopefully.

haixuanTao:
Ok, i'm fine with the current implementation for now. Shall i merge?

pixelspark:
> Ok, i'm fine with the current implementation for now.
> 
> Yeah, adding a new step is going to be a lot of dev. Let's keep it simple for now.
> 
> Shall i merge?

Agreed, let's merge this, it's complicated enough already 👍

PR #41, opened 2022-01-15
feat: accept Cast op, execute as no-op

pixelspark:


pixelspark:
This PR should probably not be merged, instead Cast should be implemented based on the work in https://github.com/haixuanTao/wonnx/pull/48

Issue #43, opened 2022-01-15
Culling identity ops does not always work properly


pixelspark:
**Describe the bug**

In the sequencer, we recently added code that removes 'identity' operations (i.e. those that only change metadata of data, not the data itself, such as `Reshape`, `Identity`, etc.). The code does this by looking at the next op and replacing the input it receives from the identity op with the input the identity op receives itself:  `input_a -> A -> output_a -> B -> output_b` becomes `input_a -> B -> output_b` by telling `B` to use `input_a` instead of `output_a`.

However, the *next* op we consider is not always the next one in the chain: a model such as `A -> B, C -> D, B + D -> E` can have order `A C B D E`. Assuming B is an identity op, when our code considers removing node B it should change node E (to point at the output of A) but it will instead look at node `D`. 

Below is an example from BERT-Squad that shows this behaviour:

<img width="291" alt="image" src="https://user-images.githubusercontent.com/181337/149634977-90d27e12-5c91-4cd0-a092-5fa7e82835b3.png">

A solution could be to look at all nodes ahead to find the one that uses. However, I think this requires a rethinking of the sequencer's fundamental assumption that the node sequence is the right unit of analysis...

PR #45, opened 2022-01-16
Make wonnx more robust by using a DAG for optimization and transformation

pixelspark:
@haixuanTao this is a work-in-progress for adding an 'intermediate representation' - i.e. a tree of nodes that we can more easily perform operations on (such as optimizations). It can also be used to fixed the 'execution order' thing (see #44).


haixuanTao:
Just for your information, I find this paper very helpful on IR optimization: https://www.google.com/url?q=https://arxiv.org/pdf/2002.03794&sa=U&ved=2ahUKEwiCmPaIkrf1AhVDyYUKHSC5ALIQFnoECAgQAg&usg=AOvVaw1RakUB9F_8KoCDSBb2dNeE

pixelspark:
@haixuanTao thanks for the reference, will look into it!

So this turned out to be almost a full rewrite of everything except the compiler. While not finished yet (I still need to integrate the optimizations made by the 'old' sequencer) it is probably a good idea to have a look at my approach. Very broadly, it works like this:

* I start building a tree of nodes, starting from a set of output nodes of the ONNX graph (currently all output nodes, but you can make a selection! if a node is not needed for calculating a specified output, it will not even be executed)
* This intermediate tree can then be optimized (currently the only optimization consists of removing `Identity` ops, but in this step I at least want to add the current optimizations from the sequencer). Note that in this stage you could detect e.g. common subexpressions or 'constant' branches (that do not depend on inputs but e.g. only on initializers) and replace those with a constant buffer. At this stage you can also do other types of analysis (currently I have one analysis that calculated the 'execution order' for a given tree which might be useful).
* Next, 'GPU resources' are attached to the tree, resulting in a new tree. A GPU resource can either be a shader (compiled using the good old compiler), an input buffer, or a 'redirection of buffers' (when e.g. a Reshape happens, you basically just want to forward the output buffer from the previous op to the next)
* At inference, the input buffers are filled and the tree is executed using a simple recursive function call (in the future it is definitely possible to parallelize execution of steps as well here). 
* Some special logic looks up the correct data for each output node (which due to 'redirection' nodes may have a different name). A special case is a model such as `A -> Identity -> B` where output B is basically just a copy of input `A`. Because of optimizations we don't actually make such a copy but instead have a node that says 'take the buffer for A when someone asks for B'. In this case however this buffer is not readable (only writable). Hence we take a different approach and simply look up the original input vector and return that. (Similar logic applies to initializers)

As you can see the approach is a bit more complicated (and I still need to clean it up lots) but hopefully you agree that it provides a lot more opportunities for further optimization. It also feels a bit less brittle than the current approach (where many parts need to align correctly). Optimizations can be described much more directly and can be more advanced. The approach also cleanly handles edge cases (such as the one described above).

haixuanTao:
Although, I agree that the current implementation has some flaws, and having the possibilities to build more reliable optimisation has clear benefit, I just want to add those review points:
- Are you sure you're talking about a tree structure and not a Directed acyclic graph(DAG)?
- With onnx-simplifier, there aren't that many: Unused nodes, Constant nodes, Identity nodes. I know that ultimately relying on onnx-simplifier is not the greatest. But, I think that those points should not necessarily be our main concerns for optimisation. 

> Allows for executing parts of a model graph (by selecting specific outputs)

This is pretty technical, are we sure that should be a feature? And, we need that feature?

> Does not rely on the fact that the ordering of nodes in the ONNX file is the actual order to execute in but calculates execution order based on the nodes that actually need to be executed.

Ordering can be an element of optimisation but I'm not sure that the default order given by the onnx file is so flawed. I think we will probably need to benchmark that to know.

pixelspark:
> Although, I agree that the current implementation has some flaws, and having the possibilities to build more reliable optimisation has clear benefit, I just want to add those review points:
> 
> * Are you sure you're talking about a tree structure and not a Directed acyclic graph(DAG)?

It is actually a DAG indeed (both [in the spec](https://github.com/onnx/onnx/blob/main/docs/IR.md#graphs) as well as in my current implementation). 

> * With onnx-simplifier, there aren't that many: Unused nodes, Constant nodes, Identity nodes. I know that ultimately relying on onnx-simplifier is not the greatest. But, I think that those points should not necessarily be our main concerns for optimisation. I also think that it's going to be a lot of effort to implement all the simplification onnx-simplifier is doing.

Well, this basically boils down to the question: what are the goals for wonnx? If you want it to be a simple execution engine for executing sequences of ONNX operators efficiently (i.e.: pre-simplfiied ONNX models), then I agree it should not attempt much optimization. 

Actually in that case I think it would be best to not even have the current optimizations in sequencer.rs at all. This ensures the code stays simple, does one thing well, and clearly leaves any optimization to other tools. What you tell wonnx to execute will be executed without any surprises. Any optimizations (either based on the current 'sequencer' or on my graph-based framework) could be provided as a separate crate (which only requires wonnx to expose the compiler in addition to the inference API, which is currently the case). 

> > Allows for executing parts of a model graph (by selecting specific outputs)
> 
> This is pretty technical, are we sure that should be a feature? And, we need that feature?

I think this has great potential for performance optimization. Consider e.g. question answering models - these typically have two inputs (one for 'the question' and one for 'the text' - output is the most likely segment of the text that contains the answer). If you want to feed the same question to different texts (or the same texts but different questions) it might be possible to precompute the values that don't change between invocations (because it only depends on one of the inputs). 

> > Does not rely on the fact that the ordering of nodes in the ONNX file is the actual order to execute in but calculates execution order based on the nodes that actually need to be executed.
> 
> Ordering can be an element of optimisation but I'm not sure that the default order given by the onnx file is so flawed. I think we will probably need to benchmark that to know.

Yeah, I checked the ONNX spec and it turns out [the nodes in a graph should be topologically sorted](https://github.com/onnx/onnx/blob/main/docs/IR.md#graphs). So you are correct that simply sequentially executing ops in the order of occurrence in the ONNX graph is valid.

Crucially though, you would need further analysis to be able to execute ops in _parallel_. If that is ever going to be desirable (and I think it is, as we cannot always saturate a GPU by just running a single op), then you will need some sort of graph representation to find out which nodes can be calculated independently.

Curious to hear your thoughts about this! I think it is very good to discuss the goals and scope of wonnx and I would be totally fine with leaving graph analysis (c.q. this PR) to separate crates (i.e. using wonnx as compiler), but I think in that case we should make this more explicit (i..e in README).

haixuanTao:
> Well, this basically boils down to the question: what are the goals for wonnx? If you want it to be a simple execution engine for executing sequences of ONNX operators efficiently (i.e.: pre-simplfiied ONNX models), then I agree it should not attempt much optimization.

> Actually in that case I think it would be best to not even have the current optimizations in sequencer.rs at all. This ensures the code stays simple, does one thing well, and clearly leaves any optimization to other tools. What you tell wonnx to execute will be executed without any surprises. Any optimizations (either based on the current 'sequencer' or on my graph-based framework) could be provided as a separate crate (which only requires wonnx to expose the compiler in addition to the inference API, which is currently the case).

I guess it is kind of what I had in mind at first. But what I figured was that there was still optimisation that could be done after tools like onnx-simplifier if we take into account the way wgpu works. That's why I built the optimiser. To me any optimisation done within wonnx should be about webgpu related optimisation. E.g. Conv + Relu . 

In this context, I think that the DAG based optimisation could make sense. compared to the current "string" match. But ultimately it should be looking for webgpu optimisation. Recycling buffers could be a great usecase as another example. 

> I think this has great potential for performance optimization. Consider e.g. question answering models - these typically have two inputs (one for 'the question' and one for 'the text' - output is the most likely segment of the text that contains the answer). If you want to feed the same question to different texts (or the same texts but different questions) it might be possible to precompute the values that don't change between invocations (because it only depends on one of the inputs).

Ok. well in that case, i would prefer to hold this feature for later if you don't mind.

> Crucially though, you would need further analysis to be able to execute ops in parallel. If that is ever going to be desirable (and I think it is, as we cannot always saturate a GPU by just running a single op), then you will need some sort of graph representation to find out which nodes can be calculated independently.

I think that parallel computing could be great, but I agree it is kind of complicated....

> Curious to hear your thoughts about this! I think it is very good to discuss the goals and scope of wonnx and I would be totally fine with leaving graph analysis (c.q. this PR) to separate crates (i.e. using wonnx as compiler), but I think in that case we should make this more explicit (i..e in README).

I think that another crate for replacing the whole onnx-simplifier could be a great idea. Even making another IR output with another Execution provider could be a great idea. I think this is not what I was planning for wonnx, as the work required is a bit much for me. 

I'm ok integrating an optimisation layer as a DAG as well, I think it makes sense as explained before. But I think we probably have to pick the minimalist feature set to avoid bugs or regressions.

pixelspark:
OK, so the current version works quite well already! There is a DAG and you can for example easily splice a node from it (for instance an 'Identity' node) and everything still works as expected. Still WIP because of two things:

* The current sequencer optimizations are not integrated yet (which is why test_resize currently fails: sequencer rewrites some inputs to attributes). 
* It is still way too complicated. For instance, the current version builds a DAG of the nodes, and then another one with GPU resources attached. The last step is largely redundant and should be replaced with a much simpler design (more akin to the current sequencer).

One thing I just realized is that the IR DAG may greatly assist in safely applying optimizations. One example: the current sequencer will always merge Conv+Relu if they are next to each other. But if the output of just the Conv is (also) used further on, it will not be accessible anymore. With the DAG this is much easier to verify (without having to worry about changing variable names etc.). 

Another idea for later is to improve buffer management. We can easily tell at which point a buffer can be disposed of because it will never be used again - this may be useful to prevent OOM when running larger models.

Still some work to do - will let you know if there's anything to review!

pixelspark:
Some other things changed by this PR:

* Do not inject the names of inputs/outputs in shader code anymore. Instead shaders use static names for inputs/outputs (they are not shared between shaders anyway). This makes the shader code cleaner and the compiler a bit less complicated.
* In the current sequencer you remove certain inputs from Resize/Reshape/Split/Clip operators and turn them into an attribute. However the current version only works if the inputs have certain exact names! in ONNX, inputs are bound positionally (for e.g. Resize, there are [four inputs](https://github.com/onnx/onnx/blob/main/docs/Operators.md#inputs-1---4): X, roi, scales and sizes). The current code would only move e.g. 'roi' to an attribute if the input had that exact name. My new version will always move the second (and next) inputs to an attribute.

Note all tests pass except for test_resize. Will have to look into this, perhaps the test is wrong. Also I still have to implement some of the optimizations (such as ConvMish) in the current sequencer. Nevertheless this is starting to look really good!

pixelspark:
Fixes #43

pixelspark:
Some other notable changes:

* The `copy.wgsl` shader has been removed (as we can now reliably 'forward buffers', so no copying required). 
* For Dropout, Squeeze, Unsqueeze, Flatten, we now forward buffers. The IR nodes still exist so the next operation knows the new shape. For Identity, the IR node is simply removed.
* For the `activation.wgsl` and `map.wgsl` shaders I now also determine a workgroup_size. The number of invocations of the shader equals threads*workgroup_size, so if you multiply workgroup_size you can divide the number of threads accordingly. This helps to avoid reaching the limits for the number of threads.

pixelspark:
Hi @haixuanTao, this is ready for your review! It's quite a big PR but I feel it is quite robust and really improves the architecture of wonnx. I hope you will consider merging it.

Please see my comments above about some peculiarities I found along the way (among other things, an optimization for Resize was implemented such that it only worked during tests with specific input names!).

The last version now also implements 'op fusing' (Conv+Relu=>ConvRelu). The following optimizations that are in the current version are not (yet) implemented, let me know your thoughts about whether they are still relevant. I think it is a better idea to first merge this without these last few optimizations (I believe that they are not that important but you may have a different opinion of course...):

* ConvMish (Conv followed by activation: `input * tanh(log(1 + exp(input))`). The sequencer recognizes this construct by looking for `["Conv", "Exp", "Add", "Log", "Tanh", "Mul", ..]`. The DAG-based optimizer only allows 'chain' optimizations of series of operators that have single inputs. It should nevertheless be possible to implement this later.
* The padding optimization for Conv. Not sure what this is doing?

````rust
if let Some(data) = initializers.get(input) {
                    let data = if input == &inputs[1]
                        && get_attribute::<Vec<i64>>("kernel_shape", None, &nodes[0])? == [3, 3]
                        && (get_attribute("pads", Some(vec![0, 0, 0, 0]), &nodes[0])?
                            == [1, 1, 1, 1]
                            || get_attribute(
                                "auto_pad",
                                Some("SAME_UPPER".to_string()),
                                &nodes[0],
                            )? == "SAME_UPPER")
                        && get_attribute("strides", Some(vec![1, 1]), &nodes[0])? == [1, 1]
                    {
                        padding(data, 12, 4)
                        // data.to_vec()
                    } else {
                        data.to_vec()
                    };
                    let data = data.to_vec();
````

* The Mul/Add optimizations in the current sequencer. Might want to generalize this a bit more before re-implementing.

**Comment by haixuanTao** [comment]
I had started to have a quick look at this PR and when I try to run the squeezenet model I don't get the expected pelican result:

```bash
cargo run --example squeeze --release       
```

haixuanTao:
Ok. so I think it is linked to the fact that you removed:
```
if let Some(data) = initializers.get(input) {
                    let data = if input == &inputs[1]
                        && get_attribute::<Vec<i64>>("kernel_shape", None, &nodes[0])? == [3, 3]
                        && (get_attribute("pads", Some(vec![0, 0, 0, 0]), &nodes[0])?
                            == [1, 1, 1, 1]
                            || get_attribute(
                                "auto_pad",
                                Some("SAME_UPPER".to_string()),
                                &nodes[0],
                            )? == "SAME_UPPER")
                        && get_attribute("strides", Some(vec![1, 1]), &nodes[0])? == [1, 1]
                    {
                        padding(data, 12, 4)
                        // data.to_vec()
                    } else {
                        data.to_vec()
                    };
                    let data = data.to_vec();
```

This piece of code put some padding to convolution with kernel 3x3. It makes the computation matrixable and increase the performance. Without the padding the shader `pool/conv_kernel_3.wgsl` is unusable.

- We should either reintroduce the logic above to check for performance regression. 
- Or accept and remove  `pool/conv_kernel_3.wgsl`  but the code is going to be slower for squeezenet.

src/compiler.rs
```diff
@@ -147,91 +143,57 @@ pub enum CompileError {
...
     },
+
+    #[error("the model exceeds the limit for {0}: {1} > {2}")]
+    ComputeLimitExceeded(String, u32, u32),
 }
 
 pub fn compile(
     node: &crate::onnx::NodeProto,
-    shape_infos: &HashMap<String, Shape>,
+    input_shapes: &[&Shape],
```
haixuanTao, line 153: I tend to prefer HashMap as it is either to manage name than stack overflow. 

Have also in mind that input and output can be optional like: https://github.com/onnx/onnx/blob/main/docs/Operators.md#batchnormalization 

What do you think?

pixelspark: @haixuanTao if I understand the specification correctly, inputs in ONNX are 'positional', meaning that their position in the list of inputs of a node determines what they are used for. For example, in a 'Reshape' operation, the first input (0) is always the input tensor and the second input (1) is always the 'shape' input, regardless of their names.

For optional inputs, the name is an empty string (I am not sure what shape onnx-simplifier assigns in this case, probably an empty shape). Strictly speaking `input_shapes: &[Option<&Shape>]` would be better I guess.

The *names* of the inputs (in an ONNX file) refer to the outputs (or initializer or inference input) that the input data should be obtained from. They are used to form the DAG, effectively. The compiler itself doesn't need to know the names. In fact it should only ever distinguish inputs by their position. 

Because of this I found it to be much cleaner to not let the compiler even see the names. Also in the DAG, names are not used anymore (well except for inference input/outputs) because otherwise we would have to rename inputs when we rename/replace outputs.

pixelspark:
> By removing `pool/conv_kernel_3.wgsl` from the compiler the code give the expected result.

Thanks for figuring this out!

> * We should either reintroduce the logic above to check for performance regression.
> * Or accept and remove  `pool/conv_kernel_3.wgsl`  but the code is going to be slower for squeezenet.

OK, I will look into this.

pixelspark:
OK, I [re-added the padding optimization](https://github.com/haixuanTao/wonnx/pull/45/commits/d990ca2b6ee2cb7b646a0b69f6067ac32d0e4bf6) and the test seems to pass (at least on my computer) 🎉 Also changed the `#[ignore]` on the test to `#[cfg(not(target_os="windows"))]` so the test runs on the other platforms (I am on a Mac)

pixelspark:
OK, the squeeze model works fine now, but it is a lot slower than current master branch (~6 seconds vs. 513 ms). Will look into this later. I think the optimizer visiting nodes more often than it should.

pixelspark:
@haixuanTao the last commit fixes the issue with the squeeze example (my suspicion was right; because the DAG 'converges' in SqueezeNet the optimizer optimized the same nodes multiple times). 

image

From my perspective this PR is ready to be merged, but feel free to comment!

haixuanTao:
Ok, I have reviewed all of it and I approved the merge. I think it is a great addition!

Thanks for the PR.

haixuanTao:
> @haixuanTao the last commit fixes the issue with the squeeze example (my suspicion was right; because the DAG 'converges' in SqueezeNet the optimizer optimized the same nodes multiple times).
> 
> image
> 
> From my perspective this PR is ready to be merged, but feel free to comment!

I think it is really interesting to see that the prediction time is the same on your GPU (M1 chip) as on mine (old GTX 1050) which mean that the 25ms, is not tied to the GPU performance. 

I have tried to do some profiling of the time spent for squeezenet but haven't found the bottleneck. The 25ms could actually be CPU-bound or IO-bound. I think #47 could be a great test to see if recycling buffer could reduce this bottleneck.

haixuanTao:
I will migrate to https://github.com/pixelspark/nnx to make more accurate benchmark.

pixelspark:
> I think it is really interesting to see that the prediction time is the same on your GPU (M1 chip) as on mine (old GTX 1050) which mean that the 25ms, is not tied to the GPU performance.

Mine is actually an M1 Max with 32 GPU cores (the original M1 has only 8 or even 7 in certain MacBooks). It should normally run circles around your GTX 1050 ;-)

Looking at the GPU usage I suspect we are nowhere near saturating it and most of the time is spent in overhead. Perhaps having less buffers can decrease this further. 25ms doesn't sound bad but is 'only' 40fps (suppose you'd want to use this on real-time camera images for instance), so there should still be room for optimization!

As another data point, my old 2014 MacBook (Intel i7, nVidia GT750M as well as Iris Pro iGPU) takes about 120ms per inference (on the Squeeze example). Interestingly it selects the iGPU instead of the dGPU. We should probably add some way to select integrated or discrete graphics.

> I have tried to do some profiling of the time spent for squeezenet but haven't found the bottleneck. The 25ms could actually be CPU-bound or IO-bound. I think #47 could be a great test to see if recycling buffer could reduce this bottleneck.

How are you profiling? I tried `cargo flamegraph` but it is not very informative either...

haixuanTao:
interesting.

So, at the moment, I'm just looking at the log with `RUST_LOG=debug` and checking the time stamp at the micro scale: https://docs.rs/env_logger/latest/env_logger/struct.Builder.html#method.format_timestamp_micros

But there is no information that is been produced after the dispatch call... There is https://github.com/Wumpf/wgpu-profiler but I did not succeed to make it work. (Didn't try very hard)

PR #46, opened 2022-01-22
Feature/python binding 

haixuanTao:
- Add another crate for python binding
- Create a Github workflow to push precompilation to pypi
- Layout groundwork for backend conformity test although most tests are not passing at the moment.


py-wonnx/src/lib.rs
```diff
@@ -0,0 +1,43 @@
+use pyo3::prelude::*;
+use pyo3::types::PyDict;
+use std::collections::HashMap;
+
+use wonnx::Session;
+#[pyclass]
+#[repr(transparent)]
+pub struct PySession {
```
pixelspark, line 8: Could make this class name simply 'Session' in Python by using `#[pyclass(name="Session")]` (see [here](https://pyo3.rs/master/class.html#customizing-the-class)) :-)

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
```
pixelspark, line 1: After this PR the main `wonnx` package is in the root whereas the python stuff is in a subdirectory. A more 'rusty' way of organising the repository is using a [workspace](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html). This basically means you have a directory for 'wonnx' (with its Cargo.toml), a directory for 'wonnx-python' (with its Cargo.toml) and a Cargo.toml in the root specifying 'wonnx' and 'wonnx-python' as workspace members.

py-wonnx/src/lib.rs
haixuanTao: Yeah, the thing is I may wrap PySession within a Python Class called Session, to add other functionality, sort of like in polars: https://github.com/pola-rs/polars/blob/master/py-polars/polars/internals/frame.py

Thanks for the suggestion though. Next time I'll run the PR by you.

Cargo.toml
haixuanTao: Yes, you're right. I'll fix that later.

PR #48, opened 2022-02-06
Support data types other than f32

pixelspark:
This PR (based on my earlier PR #45 that implemented the DAG) adds support for data types other than `f32`. Right now I have implemented support for `i32` inputs, but others (such as `f16`) should be fairly easy to add. The current version will always cast outputs to `Vec<f32>` but of course the output type should depend on the actual output type in the model.

Notably, BERT and some other models appear to use `int64` inputs - working with i64 in WGSL however is not (yet?) possible. We should therefore probably re-interpret these as `int32`. 

Some implementation notes:

* `Shape` now has a `data_type` field of type `ScalarType`. The latter holds some information about the type's size and striding.
* In `struct.wgsl`, I define a type `Scalar` to be the 'default' scalar type (all ops except those that do conversions work in a single type) based on template variable `scalar_type`. It also defines `Vec3`, `Vec4`, `Mat4x4` etc. which are the appropriate vector/matrix types. All shaders use these type names instead of the raw ones.
* The compiler sets `scalar_type` variable by looking at the types of inputs/outputs. They should all agree (except in cases where there may be inputs/outputs of different type, which happens e.g. for `Reshape` where the first input is the data and the second input is a shape, with integer data type) or else an error is thrown.
* Input tensors are now wrapped in an enum, e.g. `InputTensor::F32(my_input.as_slice())`. Output tensors are currently still all of type `Vec<f32>` (when the actual output data type is different the current version casts the elements to `f32` first).

See `tests/arithmetic.rs` for an example of integer addition.


pixelspark:
@haixuanTao I rebased this on top of your workspace PR, this one should be ready to go now.

README.md
```diff
@@ -83,7 +83,7 @@ cargo run --example mnist --release
 fn main() -> HashMap<String, Vec<f32>> {
     let mut input_data = HashMap::new();
     let image = load_squeezenet_image(); // Load image
-    input_data.insert("data".to_string(), image.as_slice().unwrap());
+    input_data.insert("data".to_string(), InputTensor::F32(image.as_slice().unwrap()));
```
haixuanTao, line 86: Shall we not be able to do inference on the data type by using `From` `Into`, removing the need for the user to specify again the type of the tensor.

pixelspark: That would be nicer indeed (it would look like `image.as_slice().unwrap().into()`).

haixuanTao: Well I was even thinking about putting the into within the function with the input hasmap value having a signature of type generic `impl Into`. Not sure what the implications are, but seems more logical for me.

haixuanTao: > If you want to feed a model inputs of different type (e.g. one F32 tensor and one I32 tensor, which happens sometimes) then you will be back to InputTensor anyway as it allows you to pass either in the same HashMap.

That is true.

Ok, I tried generics as well and it was indeed not the best solution.

The builder pattern looks interesting. 

I'm ok to keep the API at the current then.

Good discussion

haixuanTao:
Ok I think we can merge this branch.

pixelspark:
> Ok I think we can merge this branch.

Yes, agreed. We might improve the API in a later PR (e.g. the builder pattern or something else)

PR #53, opened 2022-02-09
Add a CLI utility

pixelspark:
Some other things this PR does:

* Restructure the workspace such that models/images are at the top level and more easily usable from e.g. Python
* Add the squeezenet label set (this allows removing the dev-dependency on ureq, which pulls in a lot of other stuff, as well as the ONNX model zoo on S3).
* Make the Python module a 'non default' member of the workspace, so you can `cargo build` the workspace and not be bothered by errors from the Python module's build process (which only really works with maturin I guess).

Best to merge this after #48 has landed by the way!


wonnx-cli/src/text.rs
```diff
@@ -0,0 +1,134 @@
...
+    pub tokenizer: Tokenizer,
+}
+
+pub struct BertEncodedText {
+    pub encoding: Encoding,
+}
+
+impl BertTokenizer {
+    pub fn new(vocab_path: &Path) -> BertTokenizer {
+        let wp_builder = WordPieceBuilder::new()
```
haixuanTao, line 24: I feel that this is sort of part of a broader `preprocessing step` that we should probably migrate either to a new crate or a sub-part of the main crate. Not necessarily in this PR.

wonnx-cli/src/util.rs
```diff
@@ -0,0 +1,205 @@
+use image::imageops::FilterType;
```
haixuanTao, line 1: I think that this file should be called `utils` for utilities.

pixelspark: OK, easy fix :-)

pixelspark:
Thanks! I am on holiday, will have a look at this later. Bottom line is that the CLI should be as simple as possible (argument parsing and invocation of wonnx/tract and not much else) and any processing should be offered somewhere as an optional feature or separate package.

pixelspark:
OK @haixuanTao this one is ready to go as far as I'm concerned!

haixuanTao:
So to me everything looks fine except maybe that we lack integration test of the cli. This could be beneficial too avoid especially due to crosscompilation. But, this is already a big PR so it is up to you.

haixuanTao:
Yeah and on my machine it seems that the test do not pass due to imprecision that kind of sucks.

```
---- test_reciprocal stdout ----
thread 'test_reciprocal' panicked at 'assertion failed: `(left == right)`
  left: `[1.0, 0.5, 0.33333334, 0.25, 0.2, 0.16666667, 0.14285713, 0.125, 0.11111111, 0.1, 0.09090909, 0.083333336, 0.07692307, 0.07142857, 0.06666666, 0.0625]`,
 right: `[1.0, 0.5, 0.33333334, 0.25, 0.2, 0.16666667, 0.14285715, 0.125, 0.11111111, 0.1, 0.09090909, 0.083333336, 0.07692308, 0.071428575, 0.06666667, 0.0625]`', wonnx/tests/arithmetic.rs:55:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

pixelspark:
> So to me everything looks fine except maybe that we lack integration test of the cli. This could be beneficial too avoid cross-compilation errors. and maintenance hazard in the future. But, this is already a big PR so it is up to you.

Adding a few simple tests is easy enough, and I agree they should be there. Will add a few.

pixelspark:
> Yeah and on my machine it seems that the test do not pass due to imprecision.
> 
> ```
> ---- test_reciprocal stdout ----
> thread 'test_reciprocal' panicked at 'assertion failed: `(left == right)`
>   left: `[1.0, 0.5, 0.33333334, 0.25, 0.2, 0.16666667, 0.14285713, 0.125, 0.11111111, 0.1, 0.09090909, 0.083333336, 0.07692307, 0.07142857, 0.06666666, 0.0625]`,
>  right: `[1.0, 0.5, 0.33333334, 0.25, 0.2, 0.16666667, 0.14285715, 0.125, 0.11111111, 0.1, 0.09090909, 0.083333336, 0.07692308, 0.071428575, 0.06666667, 0.0625]`', wonnx/tests/arithmetic.rs:55:5
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```

This test still does ‘exact’ comparison. For some reason that works fine on M1 but NVIDIA and Intel floats apparently are different enough to cause differences even for something as simple as ‘1/x’. Will update the test to use `approx` comparison.

pixelspark:
@haixuanTao I have added a simple CLI test and I fixed the reciprocal test to use approximate comparison!

haixuanTao:
Perfect :1st_place_medal:

PR #58, opened 2022-02-21
Implement Gather operation

pixelspark:
Implements the Gather operation, as used in e.g. BERTsquad. In a nutshell, this operator accepts a data tensor (typically f32) and an indexes tensor (typically i32), and produces a new tensor that has the elements at the specified indices from the data array. Using the 'axis' attribute it is possible to select indices from inner dimensions.

https://github.com/onnx/onnx/blob/main/docs/Operators.md#gather

Axis=0 case seems to work, need some more time to implement support for non-zero axis cases.

### DIFF
--- a/wonnx/templates/endomorphism/gather.wgsl
+++ b/wonnx/templates/endomorphism/gather.wgsl
@@ -4,23 +4,23 @@ struct Indices {
 	data: [[stride(4)]] array<i32>;
 };
 
-struct Batch {
-	data: [[stride({{ scalar_stride * batch_size }})]] array<{{ batch_type }}>;
+struct Chunk {
+	data: [[stride({{ scalar_stride * chunk_size }})]] array<{{ chunk_type }}>;
 };
 
 [[group(0), binding(0)]]
-var<storage, read> input_0: Batch; // data
+var<storage, read> input_0: Chunk; // data
 
 [[group(0), binding(1)]]
 var<storage, read> input_1: Indices; // indices
 
 [[group(0), binding(2)]]
-var<storage, write> output_0: Batch;
+var<storage, write> output_0: Chunk;
 
 [[stage(compute), workgroup_size({{ workgroup_size_x }}, {{ workgroup_size_y }})]]
 fn main([[builtin(global_invocation_id)]] global_id: vec3<u32>) {
 	let index_index = global_id.x; // Index of the index in the indices array that we are currently processing
-	let batch_index = global_id.y; // Batch of elements that we are copying for this index (batch size determined dynamically)
+	let chunk_index = global_id.y; // Chunk of elements that we are copying for this index (chunk size determined dynamically)
 
 	// Negative indexing is apparently allowed; see https://github.com/onnx/onnx/blob/main/docs/Operators.md#inputs-38
 	var index = input_1.data[index_index];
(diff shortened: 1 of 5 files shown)


wonnx/templates/endomorphism/gather.wgsl
```diff
@@ -0,0 +1,33 @@
...
+var<storage, read> input_1: Indices; // indices
+
+[[group(0), binding(2)]]
+var<storage, write> output_0: Chunk;
+
+[[stage(compute), workgroup_size({{ workgroup_size_x }}, {{ workgroup_size_y }})]]
+fn main([[builtin(global_invocation_id)]] global_id: vec3<u32>) {
+	let index_index = global_id.x; // Index of the index in the indices array that we are currently processing
+	let chunk_index = global_id.y; // Chunk of elements that we are copying for this index (chunk size determined dynamically)
+	let index_stride = u32({{ i_chunks[0][0] / chunk_size }});
```
haixuanTao, line 24: Wouldn' t that be slightly more costly than:
`let index_stride = {{ i_chunks[0][0] / chunk_size }}u;`

wonnx/templates/endomorphism/gather.wgsl
```diff
@@ -0,0 +1,33 @@
...
+
+	// Negative indexing is apparently allowed; see https://github.com/onnx/onnx/blob/main/docs/Operators.md#inputs-38
+	var index = input_1.data[index_index];
+	if(index < 0) {
+		index = {{ i_shape[0][0] }} + index;
+	}
+
+	output_0.data[(index_index * index_stride) + chunk_index] = input_0.data[(index * i32(index_stride)) + i32(chunk_index)];
```
haixuanTao, line 32: ` (index * i32(index_stride)) + i32(chunk_index)`  -> ` (index + 1) * i32(chunk_index)`  

seems less operation.

wonnx/templates/endomorphism/gather.wgsl
pixelspark: Well, a smart compiler should be able to figure out that both expressions resolve to a constant, not sure how smart this compiler is :-) I will change to your suggestion.

wonnx/templates/endomorphism/gather.wgsl
pixelspark: I'm not sure that gives the same result though? ;-)

pixelspark:
@haixuanTao fyi, the next things I will be looking at is implementing `ReduceMean` (and possibly other `Reduce*`) as well as `OneHot`. Both are needed for BERTsquad (which is what I would eventually like to be able to run).

wonnx/templates/endomorphism/gather.wgsl
haixuanTao: Yes indeed 😅 sorry!

haixuanTao:
Ok sounds great! I'm leaving my company and starting a rust position in march so I'm still taken at the moment. Will take another look at wonnx when the dust of change will have settled :)

pixelspark:
> Ok sounds great! I'm leaving my company and starting a rust position in march so I'm still taken at the moment. Will take another look at wonnx when the dust of change will have settled :)

OK, congrats!

PR #61, opened 2022-03-04
Compile wonnx to WASM, usable in WebGPU-capable browsers

pixelspark:
Still very much a work in progress, but a simple test works!


pixelspark:
@haixuanTao this works pretty well now (tested on Chrome Canary with WebGPU enabled). API looks like this at the moment:

````js
<script type="module">
import init, { Session, Input } from "/target/pkg/wonnx.js";

async function fetchBytes(url) {
	const reply = await fetch(url);
	const blob = await reply.arrayBuffer();
	const arr = new Uint8Array(blob);
	return arr;
}

async function run() {
	try {
		const [modelBytes, initResult] = await Promise.all([fetchBytes("models/single_relu.onnx"), init()])
		const session = await Session.fromBytes(modelBytes);

		const input = new Input();
		input.insert("x", [13.0, -37.0]);
		const result = await session.run(input);
		console.log({result});
		session.free();
	}
	catch(e) {
		console.error(e, e.toString());
	}
}

run();
</script>
````

**Comment by haixuanTao** [comment]
Wow! Incredible feature!
Just for your information, I have been able to run the model on Firefox Nightly.

I'll add reviews once you unmark draft from the PR.

File: README.md [hunk 1]
```diff
@@ -76,6 +76,26 @@ maturin develop
 
 Then run `python3` with the above Python code!
 
+### WASM
```
haixuanTao, line 79: I think this should be in a README inside the wonnx-wasm repository to avoid this README to be packed with to many things.

We can then add those documentation to docs.rs using:
```
#[doc = include_str!("../README.md")]
```

Only putting a light reference here.

pixelspark:
@haixuanTao So I managed to get SqueezeNet running in the browser, using the webcam as input! https://pixelspark.nl/wp-content/public/squeeze/squeeze.html 

image

wonnx/examples/mnist.rs
```diff
@@ -26,10 +25,7 @@ async fn execute_gpu() -> Option<HashMap<String, Vec<f32>>> {
     let mut input_data = HashMap::new();
 
     let image = load_image();
-    input_data.insert(
-        "Input3".to_string(),
-        InputTensor::F32(image.as_slice().unwrap()),
-    );
+    input_data.insert("Input3".to_string(), image.as_slice().unwrap().into());
```
haixuanTao, line 28: I think we have to update the README accordingly. We can maybe reference this file in order to avoid having to manually update the README

haixuanTao:
Wow! That is incredible! 

Fantastic work!

pixelspark:
> It does not work from scratch on Firefox Nightly Linux, not sure why.

Doesn't work here either in Firefox Nightly on Mac. I am seeing some warnings which might be related:

image

It works for me in Chrome Canary (on Mac). To make it work there I had to fix some issues in WGSL first (the Chrome WGSL compiler is a bit more strict than wgpu's), there might be other things that need to be fixed for Firefox.

README.md
pixelspark: Agreed, the main README should say something like "In order to use from JS, just `npm install wonnx-wasm`" (i.e. aimed at end users) whereas `wonnx-wasm/README.md` should describe the build process for developers.

pixelspark: Did not know about that include_str trick by the way!

wonnx/examples/mnist.rs
pixelspark: Agreed. (And now I am wondering if Rust will run doctests (i.e. check if the code in documentation actually runs) if you use the include_str trick...)

haixuanTao:
So, I know some of the tests are not working due to the git lfs limit. I feel like several small OSS project have migrated there static data to another provider like dropbox or else. I can open a discussion on that.

For me this PR is fine for merge.

pixelspark:
> So, I know some of the tests are not working due to the git lfs limit. I feel like several small OSS project have migrated there static data to another provider like dropbox or else. I can open a discussion on that.

This is a bit weird, as we are not using that much storage... not sure about the alternatives (Dropbox is likely traffic-limited as well?)

> For me this PR is fine for merge.

:+1: will merge!

PR #67, opened 2022-03-12
Implement Reduce{Mean,Sum,Min,Max,Prod}

pixelspark:
These are needed for BERTSQuAD at least.

(I can quite easily add the other Reduce* ops before merging this, e.g. ReduceL1, ReduceL2, ReduceSumSquare, ReduceLogSum, ReduceLogSumExp)


pixelspark:
ReduceL1, ReduceL2, ReduceSumSquare, ReduceLogSum, ReduceLogSumExp are now implemented as well!

haixuanTao:
This is amazing! And, I think, it's really great that we are reusing the template for reduction. 

Just for information, future optimization could be done using scan: https://developer.nvidia.com/gpugems/gpugems3/part-vi-gpu-computing/chapter-39-parallel-prefix-sum-scan-cuda

In the premise of this package, I have tried to implement this and it increased the speed of reduction. But, it is quite complex.

haixuanTao:
So, as you have probably seen, I've updated the python framework to test more largely our operator with the python onnx backend.
The current reduce that implemented seems too have some unhandle case:
```
FAILED tests/test_specific_op.py::OnnxBackendNodeModelTest::test_reduce_sum_default_axes_keepdims_example_cpu - pyo3_runtime.PanicExcept...
FAILED tests/test_specific_op.py::OnnxBackendNodeModelTest::test_reduce_sum_default_axes_keepdims_random_cpu - pyo3_runtime.PanicExcepti...
FAILED tests/test_specific_op.py::OnnxBackendNodeModelTest::test_reduce_sum_do_not_keepdims_example_cpu - pyo3_runtime.PanicException: c...
FAILED tests/test_specific_op.py::OnnxBackendNodeModelTest::test_reduce_sum_do_not_keepdims_random_cpu - pyo3_runtime.PanicException: ca...
FAILED tests/test_specific_op.py::OnnxBackendNodeModelTest::test_reduce_sum_empty_axes_input_noop_example_cpu - pyo3_runtime.PanicExcept...
FAILED tests/test_specific_op.py::OnnxBackendNodeModelTest::test_reduce_sum_empty_axes_input_noop_random_cpu - pyo3_runtime.PanicExcepti...
FAILED tests/test_specific_op.py::OnnxBackendNodeModelTest::test_reduce_sum_keepdims_example_cpu - pyo3_runtime.PanicException: called `...
FAILED tests/test_specific_op.py::OnnxBackendNodeModelTest::test_reduce_sum_keepdims_random_cpu - pyo3_runtime.PanicException: called `R...
FAILED tests/test_specific_op.py::OnnxBackendNodeModelTest::test_reduce_sum_negative_axes_keepdims_example_cpu - pyo3_runtime.PanicExcep...
FAILED tests/test_specific_op.py::OnnxBackendNodeModelTest::test_reduce_sum_negative_axes_keepdims_random_cpu - pyo3_runtime.PanicExcept...
```

We don't have to fix all test case. But I think, in the future, we will be able to add a flag in the readme for ops that are not 100% passing.

pixelspark:
> So, as you have probably seen, I've updated the python framework to test more largely our operator with the python onnx backend. The current reduce that implemented seems too have some unhandle case:

Yeah, so... in their infinite wisdom, the ONNX spec authors decided that ReduceSum (as of version 13) actually accepts the 'axes' list as input instead as attribute (all the other Reduce-ops just take it as attribute).

I have added an optimization that moves an axis list taken from an initializer to an attribute, and added a specific test case for it. Nevertheless, the ONNX test case still fails because it does not initialize the `axes` input with an initializer, but an actual input (i.e. it is dynamic, hence we'd need to re-compile the shader on each inference because the axis list could change!). I expect models in the wild to not use this feature.

I suggest we disable the ONNX test cases for reducesum for now.

haixuanTao:
I see! Yeah, I was also kind of frustrated by the mixing of axis with input... Is there any way we can maybe add an error mention the fact that this is not implemented? By checking the initializer or the input?

pixelspark:
> I see! Yeah, I was also kind of frustrated by the mixing of axis with input... Is there any way we can maybe add an error mention the fact that this is not implemented? By checking the initializer or the input?

This is already implemented! The optimizer will try to convert the axis input to an attribute. It expects the input to be an initializer but if it isn't, it will return an error: https://github.com/webonnx/wonnx/blob/02e9b0a2f4d20ca2dad35532d5687d50c9cc59e9/wonnx/src/optimizer.rs#L395

Other than that I should probably mention it in README.

haixuanTao:
I see. In that case, we can merge :)

PR #68, opened 2022-03-13
Implement OneHot operator

pixelspark:
Implementation for the case where axis=-1 (the default).


pixelspark:
This PR is currently based on `feature/reduce` (but could be made independent)

haixuanTao:
So at the moment all our output are f32 vectors. Python tests for OneHot are failing as they uses integer input and expect integer output. but i think that this is fine for now.

pixelspark:
> So at the moment all our output are f32 vectors. Python tests for OneHot are failing as they uses integer input and expect integer output. but i think that this is fine for now.

Agreed, we should also support non-f64 outputs (I can address in a separate PR, it will require changing the API and some fiddling with the Python/WASM bindings though).

PR #69, opened 2022-03-13
Process i64 data as i32 data

pixelspark:
WGSL does not appear to support 64-bit integers, so the best we can do (apart from building all sorts of complicated 64-bit logic based on 32 bit ints) is to use i32 instead (this also means: load i64 tensors as i32) and hope for the best (and wait for WGSL to start supporting 64 bit ints).

In practice it appears models use i64 but do not need the full range (i.e. in one of the ONNX files for BERTSQuAD, the input tokens are i64 tensors, but these values would also fit in i32).

Still need to implement the 64-to-32 conversion when loading tensors.

### DIFF
--- a/wonnx/src/gpu.rs
+++ b/wonnx/src/gpu.rs
@@ -1,9 +1,11 @@
 use std::{
     borrow::Cow,
     collections::{HashMap, HashSet},
+    convert::TryInto,
     sync::Arc,
 };
 
+use num::FromPrimitive;
 use thiserror::Error;
 use wgpu::{Buffer, BufferUsages, CommandEncoder};
 
@@ -69,6 +71,9 @@ pub enum GpuError {
 
     #[error("scalar type error: {0}")]
     ScalarType(#[from] DataTypeError),
+
+    #[error("value out of bounds")]
+    OutOfBoundsError,
 }
 
 enum InferenceOutput {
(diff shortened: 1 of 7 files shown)


pixelspark:
Current version supports reading int64-values inputs and initializers as int32 values (which is transparent as long as all the values fit into an int32...). See `test_int64_initializers` in `wonnx/tests/arithmetic.rs` for an example. Wonnx will log the appropriate warnings. This is supposed to be a temporary hack until WGSL supports int64 as data type (at that point we can implement support without changing the user-facing API).

pixelspark:
After this is merged I should do some more work to also support non-f64 outputs (including i64 outputs which are i32 outputs internally)

wonnx/src/gpu.rs
```diff
@@ -336,45 +337,82 @@ trait TensorProtoExtra {
...
-            let mut larger_raw_data = raw_data.to_vec();
-            larger_raw_data.resize(MINIMUM_BUFFER_SIZE_BYTES as _, 0);
-            resource::create_buffer_init(device, &larger_raw_data, self.get_name(), buffer_usage)
-        } else {
-            resource::create_buffer_init(device, raw_data, self.get_name(), buffer_usage)
-        })
+            ScalarType::I64 => {
+                // WGSL doesn't support 64 bit integers, so we load 64 bit tensors as 32 bit ints
+                log::warn!("initializers with int64 data type are not supported, converting into int32 initializer");
+                let ints: Vec<i32> = self.get_int64_data().iter().map(|x| *x as i32).collect();
```
haixuanTao, line 365: We should probably use TryFrom: https://doc.rust-lang.org/stable/std/convert/trait.TryFrom.html as it might give insightful debugging information.

wonnx/src/gpu.rs
```diff
@@ -551,6 +589,15 @@ impl GpuStep {
                             bytemuck::cast_slice(&resize(int_input.to_vec())),
                         );
                     }
+                    InputTensor::I64(int_input) => {
+                        log::warn!("reading int64 input as int32 (int64 is not supported for calculation but can be used as input as long as values fit in int32)");
+                        let int32_input = int_input.iter().map(|x| *x as i32).collect();
```
haixuanTao, line 594: Similar as above

haixuanTao (commented):
Great addition!

pixelspark:
@haixuanTao OK so I have now made the conversions fallible (note that `From<i64>` is not implemented for `f32` so `i64.try_into()` will not work, instead I use `f32::from_i64` from `num`.

wonnx/src/gpu.rs
haixuanTao: Is this unfaillable? (Genuine question)

pixelspark: > Is this unfaillable? (Genuine question)

It will wrap without warning:

````rust
let x: i64 = i64::MAX - 3;
let xx = x as i32;
println!("x={} xx={}", x, xx);
````

This will print `x=9223372036854775804 xx=-4`

pixelspark: @haixuanTao you're absolutely right! I changed this to use `i32::from_i64` just like in the other places (I forgot this one for some reason).

PR #76, opened 2022-03-20
Fixing small maxpool error

haixuanTao:
This is a draft template to fix maxpool error due to vectorisation error.

### DIFF
--- a/wonnx/templates/pool/aggregate.wgsl
+++ b/wonnx/templates/pool/aggregate.wgsl
@@ -10,7 +10,7 @@ var<storage, write> output_0: Array;
 fn main([[builtin(global_invocation_id)]] global_id: vec3<u32>) {
 	let gidx = global_id.x;
 
-	{% if i_shape[0][1] >= 4 and (o_lens[0] % 4) == 0 %}
+	{% if (i_shape[0][1] % 4) == 0 %}
 	if (gidx < {{ o_lens[0] / 4 }}u) {
 		let batch = gidx / {{ o_chunks[0][0] / 4 }}u; 
 		var rest = gidx % {{ o_chunks[0][0] / 4 }}u; 
@@ -65,10 +65,11 @@ fn main([[builtin(global_invocation_id)]] global_id: vec3<u32>) {
 			let index = base_index_2 + index_vec * {{ o_chunks[0][1] }}u;
 			output_0.data[index] = result[index_vec];
 		}
-	}
+	} else {
+	if ((gidx >= 4 * {{ o_lens[0] }}u) && ( gidx < {{ o_lens[0] }}u)) {
 	{% else %}
-
-	if (gidx < {{ o_lens[0] }}u) {
+	if ( gidx < {{ o_lens[0] }}u ) {
+	{% endif %}
 		let batch = gidx / {{ o_chunks[0][0] }}u; 
 		var rest = gidx % {{ o_chunks[0][0] }}u; 
 
@@ -112,5 +113,4 @@ fn main([[builtin(global_invocation_id)]] global_id: vec3<u32>) {
 
 		output_0.data[gidx] = result;
 	}
-	{% endif %}
 }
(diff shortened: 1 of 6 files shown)


pixelspark:
@haixuanTao it would be good idea to add a test case for this I guess. 

Other than that it looks good to me, but maybe @zimond should verify it fixes their issue.

wonnx/templates/pool/aggregate.wgsl
```diff
@@ -65,10 +65,11 @@ fn main([[builtin(global_invocation_id)]] global_id: vec3<u32>) {
 			let index = base_index_2 + index_vec * {{ o_chunks[0][1] }}u;
 			output_0.data[index] = result[index_vec];
 		}
-	}
+	} else {
+	if ((gidx >= 4 * {{ o_lens[0] }}u) && ( gidx < {{ o_lens[0] }}u)) {
```
zimond, line 69: Seems it should be {{ 4 * o_lens[0]}}u

zimond:
After this patch, my model is yielding roughly 1/2 of the right result... I will try to dig deeper to find what is happening

haixuanTao:
Ok! There are still problems with this shader. I'll investigate also this evening. Sorry if it is blocking.

zimond:
Never mind! Thanks for your time and hard work!

haixuanTao:
Ok, so one interesting mistake was that I initiated the result at 0 where some maximum could be negative. You live you learn.
So, as of now, all onnx backend test for MaxPool, AveragePool and GlobalAveragePool are passing as the CI is passing the onnx_backend_test.py 

Hope it helps you implement your model.

wonnx/templates/pool/aggregate.wgsl
```diff
@@ -56,5 +75,66 @@ fn main([[builtin(global_invocation_id)]] global_id: vec3<u32>) {
 			let index = base_index_2 + index_vec * {{ o_chunks[0][1] }}u;
 			output_0.data[index] = result[index_vec];
 		}
+	} else {
+	if ((gidx >= 4u * {{ o_lens[0] }}u) && ( gidx < {{ o_lens[0] }}u)) {
```
zimond, line 79: how could gidx >= 4 * o_lens[0] and < olens[0] at the same time?

zimond, line 79: This condition seems impossible

haixuanTao:
There might be some issue with the Add as well. I have implemented the Add as piecewise Add and other Add will not work.

wonnx/templates/pool/aggregate.wgsl
```diff
@@ -56,5 +75,66 @@ fn main([[builtin(global_invocation_id)]] global_id: vec3<u32>) {
 			let index = base_index_2 + index_vec * {{ o_chunks[0][1] }}u;
 			output_0.data[index] = result[index_vec];
 		}
+	} else {
+	if ((gidx >= 4u * {{ o_lens[0] / 4 }}u) && ( gidx < {{ o_lens[0] }}u)) {
```
zimond, line 79: I don't get it, doesn't this just mean `gidx >= o_lens[0] && gidx < o_lens[0]` which is still not possible? @haixuanTao

zimond:
> There might be some issue with the Add as well. I have implemented the Add as piecewise Add and other Add will not work.

`Add` feeds the same data to the custom op & conv. The custom op's first input being correct indicates that `Add` is correct. So it must be `conv`. I'm still digging.

By the way, I just write a version of arithmetic.wgsl which supports broadcast. It doesn't play well with the vectorization design. I think the whole vectorization optimization code need some changes like aggregate did.

zimond:
Even redirecting all kernel=(3,3) conv from `conv_kernel_3.wgsl` to `conv.wgsl` leads to different output...

wonnx/templates/pool/aggregate.wgsl
haixuanTao: Well it is an euclidean division. So it is between `[ o_lens - o_lens % 4, o_lens ]`

haixuanTao:
Just for your information @zimond , the computation is not extremely precise, there is some randomness. This is why you will find some assert_almost_equal in the codebase.

PR #77, opened 2022-03-20
Feature/integer outputs

pixelspark:
This PR implements support for output types other than Vec<f32>. 

## Implementation details

Output tensors are wrapped in an enum `OutputTensor` (e.g. `OutputTensor::F32(Vec<f32>)`). The difference with `InputTensor<'a>` is that the output tensor always owns its values. It defines a convenience method `.as_f32()` to convert to `Vec<f32>` (propagating any conversion errors) as well as a `.unwrap_f32_slice()` that supplies an `&[f32]` for tensors that you know will be of f32 type. For JS/WASM and Python, the OutputTensor is converted using the appropriate methods.

This is based on #69, so perhaps merge that one first.

### DIFF
--- a/wonnx/src/utils.rs
+++ b/wonnx/src/utils.rs
@@ -106,18 +106,10 @@ pub enum OutputTensor {
     I64(Vec<i64>),
 }
 
-impl OutputTensor {
-    pub fn unwrap_f32_slice(&self) -> &[f32] {
-        match self {
-            OutputTensor::F32(fs) => fs.as_slice(),
-            OutputTensor::I32(_) | OutputTensor::I64(_) => panic!("cannot convert into f32 slice"),
-        }
-    }
-}
-
 impl TryFrom<OutputTensor> for Vec<f32> {
     type Error = TensorConversionError;
 
+    /// Convert OutputTensor into a Vec<f32>, possibly converting integer tensors if the values fit
     fn try_from(value: OutputTensor) -> Result<Self, Self::Error> {
         match value {
             OutputTensor::F32(floats) => Ok(floats),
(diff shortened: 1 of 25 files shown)


wonnx/src/utils.rs
```diff
@@ -79,6 +82,59 @@ impl<'a> From<&'a [i32]> for InputTensor<'a> {
...
+                .map(|i| f32::from_i32(*i).ok_or(TensorConversionError::OutOfBoundsError))
+                .collect::<Result<_, _>>(),
+            OutputTensor::I64(ints) => ints
+                .into_iter()
+                .map(|i| f32::from_i64(i).ok_or(TensorConversionError::OutOfBoundsError))
+                .collect::<Result<_, _>>(),
+        }
+    }
+
+    pub fn unwrap_f32_slice(&self) -> &[f32] {
```
haixuanTao, line 120: it seems to me that `unwrap_f32_slice()` is equal to `as_f32().unwrap().as_slice()` which is more readable from my point of view. I feel like as this trait is going to be client-facing and therefore will have to be maintained. I will probably prefer to keep it as the standard library. What do you think?

haixuanTao: I know that as_f32 might be more permissive compared to unwrap_f32_slice but, this permissiveness might be a blessing if we need to handle the type of the vector with a different type for wgsl reason.

haixuanTao: I guess that we can even implement TryFrom/TryInto trait for OutputTensor for f32 / i32 / i64. I feel like it would be pretty neat. But open for discussion :)

pixelspark: Agreed, this should be implemented through `try_into` (I just committed a change implementing TryFrom, which is apparently the idiomatic way to do this). 

Note that some extra work was needed to support slices; just having `TryFrom<OutputTensor> for Vec<f32>` doesn't make them work automatically:

````rust
common::assert_eq_vector(
  result["Y"].try_into().unwrap().as_slice(), // Fails to compile: try_into needs type annotations
  some_vec.as_slice(),
);

// The canonical way to annotate would be:
common::assert_eq_vector(
  TryInto::<Vec<f32>>::try_into(result["Y"]).unwrap().as_slice(), // works but ugly
  some_vec.as_slice(),
);

````

Because the above is quite ugly, I therefore also implement a `TryFrom<&'a OutputTensor> for &'a [f32]`. This way, `result["Y"].unwrap_f32_slice()` becomes `&(result["Y"]).try_into().unwrap()`. The downsize is that this try_into only succeeds if the output tensor is already `OutputTensor::F32` (it can't do conversion from I32/I64 like the other `TryFrom for Vec<f32>`, because as it operates on references it has nowhere to store the result). This is a bit asymmetric but still the best solution as far as I can see.

PR #82, opened 2022-03-27
Implement Softmax for arbitrary axes

pixelspark:
The current implementation of Softmax only works for the last axis (which is the default case).

This PR implements Softmax for all axes. It specifically adresses the issue that Softmax-13 behaves significantly differently than previous versions (the old ones calculate the max/sum for all values in the axis 'chunk', the new version does it 'per layer'). 

Note: I added a few specific ONNX backend test cases to the Python script. However, for SoftMax there are also 'expanded' test cases which appear to be the softmax test case followed by some functions applied. These do not pass for some reason. However, the 'specific' Softmax cases do pass, and so do the test cases that I based on [the test cases in onnxruntime](https://github.com/microsoft/onnxruntime/blob/master/onnxruntime/test/providers/cpu/math/softmax_test.cc).


wonnx/src/compiler.rs
```diff
@@ -460,28 +459,65 @@ pub fn compile(
...
-                    op: "Softmax".to_string(),
-                });
-            }
+            let left_of_axis = input_shapes[0].dims[0..(axis as usize)]
+                .iter()
+                .product::<u64>();
+            let axis_chunk = input_shapes[0].dims[(axis as usize)..]
+                .iter()
+                .product::<u64>();
+            let right_of_axis_chunk = input_shapes[0].dims[((axis + 1) as usize)..]
```
haixuanTao, line 476: So in the case that the axis is the last axis. Wouldn't that overflow?

haixuanTao: Ok. So I just checked and it won't because if axis +1 = rank(). this result in a empty list which product is 1.

haixuanTao:
This seems to work. Feel free to merge.

PR #85, opened 2022-03-27
Fix readability of buffers

pixelspark:
If there is an op that forwards an input buffer without changing it (e.g. `Reshape`, `Squeeze`, etc.), the 'readability' flag should also be forwarded. Otherwise, the output will not be readable. This PR fixes that behaviour.

This occurs e.g. in BERTSQuAD, where the last two nodes are `Split --> Unsqueeze`. The output of the `Unsqueeze` is marked as 'readable output'. Normally this would cause the output buffer of the op to be created with `MAP_READ` BufferUsage so it can be read. However, because Unsqueeze only alters shape, it simply forwards the output of `Split`. Before this PR, this output would not be marked as 'must be readable' and so not created with `MAP_READ`, leading to an error when reading the results.

### DIFF
--- a/wonnx/src/gpu.rs
+++ b/wonnx/src/gpu.rs
@@ -160,7 +160,7 @@ impl GpuModel {
         nodes_seen: &mut HashSet<NodeIdentifier<'model>>,
     ) -> Result<(), GpuError> {
         let node_identifier = node.identifier();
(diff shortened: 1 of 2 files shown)


haixuanTao:
Make sense. Ok for me.

PR #86, opened 2022-03-28
Fix/mnist  Fixing misconfigured Padding 

haixuanTao:
Maxpool fix introduced errors in the value of the padding of all convolution layer that this commit fix by putting the right axes to compute the padding.

I have also changed the CI/CD in order to fix git lfs issues and not have failed test happens in the future.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -162,7 +162,7 @@ jobs:
         shell: bash
         run: |
           # Download model and image data 
-          wget https://wonnx.s3.eu-west-3.amazonaws.com/data.zip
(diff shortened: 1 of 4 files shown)


haixuanTao:
I have purposely left git lfs reference as there is a lot of Pull Request Opened and removing git lfs might be tiresome when moving between branch for developpement. 

We could later remove git lfs reference if need be.

pixelspark:
This appears to fix the mnist test case👍  I am also fine with having both LFS and an S3 download option. The LFS should probably be kept as 'authoritative' source. Feel free to merge!

pixelspark (approved):
:+1:

Issue #56, opened 2022-02-16
Update WGSL syntax when wgpu ships naga 0.9


pixelspark:
**Describe the bug**

The WGSL syntax will change with the next release of wgpu, see here: https://github.com/gfx-rs/naga/blob/master/CHANGELOG.md#v09-tbd. This will require us to rewrite shader code slightly to conform to new syntax.

pixelspark:
On Chrome Nightly, the compiler warns about the old syntax, but still happily accepts it. Not sure about Firefox...

PR #88, opened 2022-03-29
fix: upgrade to newer WGSL syntax

pixelspark:
First go at #56 (part of an attempt to fix WASM demo that appears to have stopped working in recent browser versions..).

This works when you change the wgpu dependency to `wgpu = { git="https://github.com/gfx-rs/wgpu.git" }`. This can be merged after wgpu releases 0.13.0 and we change our wgpu dependency to that version.


pixelspark:
I can move forward on this now since ship had a new release: https://github.com/gfx-rs/wgpu/releases/tag/v0.13.0

PR #89, opened 2022-04-02
feat: add a way to only partially execute a model by manually specifying output names

pixelspark:
Currently, it is only possible to construct an inference session that infers all of a model's outputs. This is not always desired (e.g. some outputs may be unused). For debugging purposes it may be useful to be able to calculate the output of a specific (intermediate) node.

This patch allows users to optionally specify a set of output names. When specified, only those outputs will be calculated and returned. Due to the way our IR is constructed, this also means that intermediate nodes that are not necessary to calculate the specified outputs will not be processed.

This adds `Session::from_{path|bytes|model}_with_options` and a `SessionOptions` struct. We can add more options to this struct in the future (e.g. GPU selection and such). Users should probably always add in a `..Default::defaults()` but I am not sure we can enforce this. We might also use a builder pattern for this. I left the old methods (`Session::from_{path|bytes|model}`) in place.

The CLI by default returns a single output (the first or the one specified by `--output-name`). It is useful to retain the 'fully calculate a model' capability, so I added a `--partial` flags that (combined with `--output_name`) will lead to partial execution up to the specified output only (in the future `--partial` could be made the default).

### DIFF
--- a/wonnx-cli/src/gpu.rs
+++ b/wonnx-cli/src/gpu.rs
@@ -1,6 +1,7 @@
 use std::collections::HashMap;
 
 use wonnx::onnx::ModelProto;
+use wonnx::SessionOptions;
 
 use async_trait::async_trait;
 use wonnx::utils::OutputTensor;
@@ -14,9 +15,14 @@ pub struct GPUInferer {
 }
 
 impl GPUInferer {
-    pub async fn new(model_path: &str) -> Result<GPUInferer, NNXError> {
+    pub async fn new(
+        model_path: &str,
+        outputs: Option<Vec<String>>,
+    ) -> Result<GPUInferer, NNXError> {
+        let session_opts = SessionOptions { outputs };
+
         Ok(GPUInferer {
-            session: wonnx::Session::from_path(model_path).await?,
+            session: wonnx::Session::from_path_with_options(model_path, &session_opts).await?,
         })
     }
 }
--- a/wonnx-cli/src/gpu.rs
+++ b/wonnx-cli/src/gpu.rs
@@ -1,7 +1,7 @@
 use std::collections::HashMap;
 
 use wonnx::onnx::ModelProto;
-use wonnx::SessionOptions;
+use wonnx::SessionConfig;
 
 use async_trait::async_trait;
 use wonnx::utils::OutputTensor;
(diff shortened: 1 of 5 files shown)


SludgePhD:
> Users should probably always add in a `..Default::defaults()` but I am not sure we can enforce this.

I don't think you can *quite* enforce this, but you can mark the struct as `#[non_exhaustive]`, which disallows creating it using the struct literal syntax. This forces users to create the struct using `Default::default()` and then they'd manually set the fields they want to be different.

pixelspark:
> > Users should probably always add in a `..Default::defaults()` but I am not sure we can enforce this.
> 
> I don't think you can _quite_ enforce this, but you can mark the struct as `#[non_exhaustive]`, which disallows creating it using the struct literal syntax. This forces users to create the struct using `Default::default()` and then they'd manually set the fields they want to be different.

Thanks for the suggestion to use `#[non_exhaustive]`! The downside is that the following is not allowed when SessionOptions is marked `#[non_exhaustive]`:
````rust
 let session_opts = SessionOptions {
            outputs,
            ..Default::default()
        };
````

This means users would indeed have to create a mutable SessionOptions object and change fields, which is not ideal (a builder pattern would be more ergonomic, e.g. something like `SessionOptions::new().with_outputs(outputs).with_other_option(x)`)

**Comment by haixuanTao** [comment]
What about implementing a method that set an optional configuration field. It will not exactly be a builder pattern but it would be pretty close and reusable in case we want to switch it to a builder pattern. I feel that is also more readable.
```rust
Session::from_{path|bytes|model}(...).with_options(...)
``` 
I would be prefers the term `config` instead of `options` that is more generic within the rust ecosystem,

pixelspark:
> What about implementing a method that set an optional configuration field. It will not exactly be a builder pattern but it would be pretty close and reusable in case we want to switch it to a builder pattern. I feel that is also more readable.
> 
> ```rust
> Session::from_{path|bytes|model}(...).with_options(...)
> ```
> 
> I would be prefers the term `config` instead of `options` that is more generic within the rust ecosystem,

Problem is, we want to know the options already when running the from_X method. Otherwise all users would have to do something like `Session::from_X().build()` so we know there won't be any options coming (*or* the whole from_X work is deferred to the first call to `run`, which is also undesirable).

Using `config` instead of `options` is fine for me by the way, will change it :+1:

haixuanTao:
Yep indeed! 

Ok well in that case the current design is ok for me. I think the builder pattern is probably better but might be tiresome to change at this stage.

PR #90, opened 2022-04-03
Make `nnx info` deterministic

SludgePhD:
The use of `HashMap` and `HashSet` made the output of `nnx info <model>` nondeterministic, which makes it hard to compare it between an unoptimized and optimized version of the same model. This PR switches the relevant data structures to their deterministic `BTree` variants to fix that.


pixelspark:
Thanks :+1: looks good to me! The CI fail is unrelated to your PR, so will merge.

PR #95, opened 2022-04-24
Gemm/MatMul improvements: support small matrices, stacking and bias broadcasting

pixelspark:
First attempt at fixing #94. 

This one does not work just right yet however because it is still using Mat4x4 for multiplication, which should probably Mat2x2 (when i_shape[1][1] == 2) or even Mat3x3 or Scalar... leading to intermittent blocks of NaNs all over the place in the output:

image

@haixuanTao can you have a look at this, perhaps give me some pointers? I am not too familiar with the Gemm implementation.

PR #99, opened 2022-05-01
Implement stacking for MatMul

pixelspark:
The [operator docs for MatMul](https://github.com/onnx/onnx/blob/main/docs/Operators.md#matmul) say:

````
Matrix product that behaves like numpy.matmul: https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.matmul.html
````

The numpy documentation in turn says:

````
The behavior depends on the arguments in the following way.

1. If both arguments are 2-D they are multiplied like conventional matrices.
2. If either argument is N-D, N > 2, it is treated as a stack of matrices residing in the last two indexes and broadcast accordingly.
3. If the first argument is 1-D, it is promoted to a matrix by prepending a 1 to its dimensions. After matrix multiplication the prepended 1 is removed.
4. If the second argument is 1-D, it is promoted to a matrix by appending a 1 to its dimensions. After matrix multiplication the appended 1 is removed.
````

This PR implements points 2 to 4.


pixelspark:
Depends on #95

pixelspark:
Moved into #95

PR #102, opened 2022-05-05
Use only primary wgpu backends by default

SludgePhD:
Using `Backends::all()` enables unsupported backends such as OpenGL, which results in obscure errors at least on my Linux system.

### DIFF
--- a/wonnx/src/resource.rs
+++ b/wonnx/src/resource.rs
@@ -2,8 +2,8 @@ use wgpu::{util::DeviceExt, BufferUsages};
 
 // Get a device and a queue, honoring WGPU_ADAPTER_NAME and WGPU_BACKEND environment variables
 pub async fn request_device_queue() -> (wgpu::Device, wgpu::Queue) {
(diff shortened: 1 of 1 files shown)


haixuanTao:
Thanks! I think it is a good default as well as it might be slow to enable all backends. 

You should probably open an issue on wgpu. As it shouldn't be blocking to enable OpenGL.

pixelspark:
@SludgePhD have you seen [this](https://github.com/webonnx/wonnx#gpu-selection)? You may select the backend of choice by setting an env variable, e.g. `WGPU_BACKEND=vulkan`.

SludgePhD:
> You should probably open an issue on wgpu. As it shouldn't be blocking to enable OpenGL.

Looks like my issue is https://github.com/gfx-rs/wgpu/issues/2513

> @SludgePhD have you seen [this](https://github.com/webonnx/wonnx#gpu-selection)? You may select the backend of choice by setting an env variable, e.g. `WGPU_BACKEND=vulkan`

Yeah, I just figured it shouldn't pick a broken backend by default if I don't set that.

PR #106, opened 2022-05-09
Implement `Pad`

SludgePhD:
This implements the `Pad` operator for `mode=constant` and without support for shrinking input dimension (ie. all `pads` values must be >=0).

`Pad` is used by many of the MediaPipe networks. I was able to *run* the [Face Detection](https://google.github.io/mediapipe/solutions/face_detection) network with this PR applied, but it produced incorrect results for as of yet unknown reasons.


SludgePhD:
It looks like the [`test_constant_pad`](https://github.com/onnx/onnx/blob/5b1346e30d66e4ec550f6b63c3883b258a2e8e3e/onnx/backend/test/case/node/pad.py#L42) test provides `pads` only as a graph input, so it won't work with the wonnx approach.

pixelspark:
> It looks like the [`test_constant_pad`](https://github.com/onnx/onnx/blob/5b1346e30d66e4ec550f6b63c3883b258a2e8e3e/onnx/backend/test/case/node/pad.py#L42) test provides `pads` only as a graph input, so it won't work with the wonnx approach.

Yeah, we've run into that before. We've implemented a workaround for some other ops that converts certain initializer inputs (which are static) to attributes (see https://github.com/webonnx/wonnx/blob/master/wonnx/src/optimizer.rs#L319). I would encourage including this workaround for Pad as well (and it shouldn't be too much work).

SludgePhD:
I think that only works if `pads` comes from a constant tensor within the graph, but here it's an input *to* the graph.

pixelspark:
> I think that only works if `pads` comes from a constant tensor within the graph, but here it's an input _to_ the graph.

Ah, if it's an inference input the workaround is not going to help. I will merge and add a note to the README. Thanks for your contribution!

redthing1:
So it looks like this is still missing the `reflect` mode. I would like to add it, but I'm totally new to WebGPU and Rust, though I have some experience with ONNX and a lot of experience with DL.
I was wondering if someone who is knowledgeable on this could help guide me or explain what would be necessary, so that I can send in a PR.

SludgePhD:
The official docs have an example for `Pad` with `mode=reflect` here: https://github.com/onnx/onnx/blob/main/docs/Changelog.md#Pad-19

I'd start by porting this to a wonnx test. `compiler.rs` then has to be updated to accept `mode=reflect` and pass the mode to the template engine. Then `pad.wgsl` can use the mode parameter to decide how to perform the padding.

redthing1:
Yeah, I think the `compiler.rs` side of it and the test looks okay. But I'm quite intimidated by the actual compute shader template.

SludgePhD:
Yeah, those could be documented better. A few general notes for many of the wonnx shaders:

- Each invocation computes exactly one output value. The invocation index `gidx` is the linear array index of that output value.
- The invocation index is divided by the output tensor dimension strides (`o_chunks`) to get the output *tensor* indices, which are the `d_N` variables in the shader.
- The output indices are then used, alongside the `Pad` node configuration, to compute which input index to load from, or whether this particular output index should be filled with the padding value.

Currently the shader just has a `pad` bool to tell it when to use the constant value instead of loading from the input tensor. That's not enough for the other padding modes, which need to compute an alternative input index instead.

It might also help to add a `println!` after the template is rendered to check whether the generated code makes sense.

redthing1:
Really appreciate your explanations, I'll take a look at this soonish and ask if I have more questions.

redthing1:
Alright @SludgePhD, I've done the outline for this (the stuff _other_ than the compute shader). But I really feel kind of stuck here and could really use some more detailed guidance.

PR #107, opened 2022-05-10
Promote `debug_assert`s to normal `assert`s

SludgePhD:
As far as I can tell, none of these asserts are in performance-sensitive code, so it makes sense to have them always checked to ensure that no invariants are violated.

### DIFF
--- a/wonnx/src/compiler.rs
+++ b/wonnx/src/compiler.rs
@@ -771,7 +771,12 @@ pub fn compile(
         op @ ("MaxPool" | "AveragePool" | "Conv" | "ConvRelu" | "ConvLeakyRelu" | "ConvMish"
         | "GlobalAveragePool") => {
             // TODO: Conv only support NxCxHxW for the moment.
-            assert!(input_shapes[0].rank() == 4);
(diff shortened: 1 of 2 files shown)


wonnx/src/compiler.rs
```diff
@@ -771,7 +771,7 @@ pub fn compile(
         op @ ("MaxPool" | "AveragePool" | "Conv" | "ConvRelu" | "ConvLeakyRelu" | "ConvMish"
         | "GlobalAveragePool") => {
             // TODO: Conv only support NxCxHxW for the moment.
-            debug_assert!(input_shapes[0].rank() == 4);
+            assert!(input_shapes[0].rank() == 4);
```
pixelspark, line 774: This one should not actually be an assert, instead an if check returning `Err(CompileError::InvalidInputShape{..})` when the condition is not met. The input ONNX model could simply be wrong and that should not cause a panic.

pixelspark (changes requested):
Good idea to change this. I have one minor issue with the first change, I agree the others should be asserts.

PR #108, opened 2022-05-10
Fix image loading with non-224x224 images

SludgePhD:


wonnx-preprocessing/src/image.rs
```diff
@@ -31,7 +31,7 @@ pub fn load_bw_image(
     })
 }
 
-// Loads an image as (1, w, h, 3)
+// Loads an image as (1, 3, w, h)
```
pixelspark, line 34: This seems inconsistent with what it says below: either this should be `(1, 3, h, w)`, or below should read `from_shape_fn((1, 3, width, height)` where it now says `height, width` instead.

pixelspark (changes requested):

pixelspark:
Good idea to make this more generic. Should we perhaps make the shape (`(1,3,w,h)`, (3,w,h)`, `(3,h,w)` etc.) configurable? I am not sure what shape your model is using vs. what SqueezeNet (which this was written for) expect, it might differ between models.

SludgePhD:
ONNX only works with NCHW, so only supporting `(1, 3, h, w)` should suffice (tf2onnx will however insert a `Transpose` node after the input to make the model accept NHWC inputs, unless `--inputs-as-nchw` is passed).

What *should* probably be made configurable is the expected input value range, since the MediaPipe networks don't require per-channel mean and stddev correction (and it needs values to be in range `[-1.0, 1.0]`). But this PR is sufficient to reproduce my issue in #109, so this isn't required for now.

Issue #112, opened 2022-05-14
Incorrect results for Transpose


pixelspark:
**Describe the bug**
Transpose appears to produce incorrect results for certain permutations.

**To Reproduce**

In the following test case, the following work (like in NumPy):
* Transpose perm=`0,2,1,3` followed by Transpose perm=`0,2,1,3`which should reverse the first transpose
* Transpose perm=`0,3,2,1` followed by Transpose perm=`0,3,2,1`which should reverse the first transpose

The following works in NumPy (see expected behaviour below), but fails in woonx:
* Transpose perm=`0,2,3,1` followed by Transpose perm=`0,3,1,2` which should reverse the first transpose

(Note that in this case, the first perm is not equal to the latter).

````rust

fn test_transpose_4d_perm(transpose_first: &[i64], transpose_second: &[i64]) {
    let mut input_data = HashMap::new();
    let data = (0..2 * 3 * 4).map(|x| x as f32).collect::<Vec<f32>>();
    input_data.insert("X".to_string(), data.as_slice().into());

    let x_dims = vec![1, 2, 3, 4];
    let intermediate_dims: Vec<i64> = transpose_first
        .iter()
        .map(|i| x_dims[*i as usize])
        .collect();

    // Model: X -> Transpose -> Y -> Transpose -> Z; X==Z
    let model = model(graph(
        vec![tensor("X", &x_dims)],
        vec![tensor("Z", &x_dims)],
        vec![tensor("Y", &intermediate_dims)],
        vec![],
        vec![
            node(
                vec!["X"],
                vec!["Y"],
                "Transpose",
                "Transpose",
                vec![attribute("perm", transpose_first.to_vec())],
            ),
            node(
                vec!["Y"],
                vec!["Z"],
                "Transpose",
                "Transpose",
                vec![attribute("perm", transpose_second.to_vec())],
            ),
        ],
    ));

    let session =
        pollster::block_on(wonnx::Session::from_model(model)).expect("session did not create");
    let result = pollster::block_on(session.run(&input_data)).unwrap();

    common::assert_eq_vector((&result["Z"]).try_into().unwrap(), &data);
}

/* This tests the equivalent of the following Python code:
a = np.arange(0,24).reshape((1,2,3,4));
a == a.transpose(a).transpose(inverse of a)
*/
#[test]
fn test_two_transposes_4d() {
    // a == a.transpose([0,2,1,3]).transpose([0,2,1,3])
[...]

PR #113, opened 2022-05-14
Fix incorrect Transpose output for certain choices of 'perm'

pixelspark:
Fix for #112


SludgePhD:
The python backend tests for transpose are also failing on git head, does this fix them?

pixelspark:
Yep, they all pass now, I just enabled them!

haixuanTao:
Looks good to me :)

PR #114, opened 2022-05-14
fix: prevent tanh from producing NaN values

pixelspark:
The `tanh` function in WGSL apparently produces NaN values for large (absolute) inputs. As `tanh` basically converges to 1, this PR clamps the input values to `tanh` to be between [-10, 10].

````python
>>> import math
>>> math.tanh(10)
0.9999999958776927
>>> math.tanh(-10)
-0.9999999958776927
````


haixuanTao:
Looks good to me :)

PR #115, opened 2022-05-18
Fix: split operator does not take all elements into account

pixelspark:
When doing a split of a `2x1x256` tensor into two `1x1x256` tensors I found that the second tensor was all zeroes. This was caused by the fact that the Split shader would only be started with `output_lengths[0]/256` workgroups (and hence would only copy 256 values from input) whereas there are in fact 512 values (so n_threads should be 2). This PR changes the number of threads to `input_lengths[0]/256` which makes this work properly.

Note we can't enable the ONNX backend tests for this op because those use dynamic inputs for the `split` input (which we don't support at the moment). The `test_split` case doesn't catch this error as the number of elements is below 256, and we simply start one thread (with workgroup size 256) in that case.


haixuanTao:
I don't see any issue with this PR :)

PR #117, opened 2022-06-05
Feature/qa improvements

pixelspark:
This PR improves some things around working with Q/A models (primarily in the preprocessing crate and the CLI):

- Return the 'answer text' based on the original input context text
- Do not set a fixed BERT specific tokenizer, instead require user to load it from a config JSON (which appears to be the typical way to do it for e.g. Huggingface transformers as well)
- Support segmenting the input in case the context text is larger than the maximum sequence size

### DIFF
--- a/wonnx-preprocessing/src/text.rs
+++ b/wonnx-preprocessing/src/text.rs
@@ -178,6 +178,52 @@ impl BertEncodedText {
             .map(|x| *x as i64)
             .collect()
     }
+
+    pub fn get_answer(&self, start_output: &[f32], end_output: &[f32]) -> String {
+        let mut best_start_logit = f32::MIN;
+        let mut best_start_idx: usize = 0;
+
+        let input_tokens = self.encoding.get_tokens();
+
+        for (start_idx, start_logit) in start_output.iter().enumerate() {
+            if start_idx > input_tokens.len() - 1 {
+                break;
+            }
+            match input_tokens[start_idx].as_str() {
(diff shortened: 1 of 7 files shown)


haixuanTao:
Please remove the bert tokenizer JSON. You can put it as a downloaded file. And, escape it, with a .gitignore

pixelspark:
> Please remove the bert tokenizer JSON. You can put it as a downloaded file. And, escape it, with a .gitignore

Oops, that shouldn't be in there. Just amended the commit.

pixelspark:
Renamed some variables and changed the default JSON path now too.

wonnx-preprocessing/src/text.rs
```diff
@@ -178,6 +128,75 @@ impl BertEncodedText {
...
+        let mut best_start_idx: usize = 0;
+
+        let input_tokens = self.encoding.get_tokens();
+
+        for (start_idx, start_logit) in start_output.iter().enumerate() {
+            if start_idx > input_tokens.len() - 1 {
+                break;
+            }
+            match input_tokens[start_idx].as_str() {
+                "[CLS]" | "[SEP]" | "[PAD]" => continue,
```
haixuanTao, line 143: Could those tokens be constant variables?

pixelspark: Good idea, I now use the `special_tokens_mask` from the tokenizer to skip special tokens (which means it will skip any special token configured for the tokenizer in e.g. the JSON).

haixuanTao:
Looks good to me thanks :)

PR #119, opened 2022-07-03
Fix/upgrade to wgpu 0.13.1

pixelspark:
This PR upgrades wgsl from v0.12.0 to v.0.13.1. This is a long awaited update that (among other things) updates the WGSL syntax to follow the latest standard (which e.g. Google Chrome (Tint) currently uses as well).

Tests pass on my machine and all examples work. Will do some more testing, there may be some edge cases here and there (in particular for ops where scalar_type is an integer - we can't use matrixes anymore for these because WGSL now only supports float-based matrixes).

### DIFF
--- a/wonnx/src/gpu.rs
+++ b/wonnx/src/gpu.rs
@@ -672,9 +672,9 @@ impl GpuTensor {
 
             wgpu::util::DownloadBuffer::read_buffer(device, queue, &buffer_slice, move |b| {
                 // Called on download completed
-                tx.send(match b {
-                    Ok(b) => Ok(Self::read_bytes_to_vec(&b, shape)),
-                    Err(abe) => Err(GpuError::BufferAsyncError(abe)),
+                tx.send(match buffer {
+                    Ok(bytes) => Ok(Self::read_bytes_to_vec(&bytes, shape)),
+                    Err(error) => Err(GpuError::BufferAsyncError(error)),
                 })
                 .unwrap();
             });
--- a/wonnx/src/gpu.rs
+++ b/wonnx/src/gpu.rs
@@ -5,9 +5,10 @@ use std::{
     sync::Arc,
 };
 
+use bytemuck::NoUninit;
 use num::FromPrimitive;
 use thiserror::Error;
-use wgpu::{Buffer, BufferUsages, CommandEncoder};
+use wgpu::{Buffer, BufferAsyncError, BufferUsages, CommandEncoder};
 
 use crate::{
     compiler::{compile, CompileError, CompiledNode},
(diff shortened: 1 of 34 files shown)


pixelspark:
And with the last few commits, WASM works again (at least on Chrome - for some reason Firefox says it cannot find a GPU given preferences on my machine, YMMV)

pixelspark:
On Firefox (Nightly `104.0a1 (2022-07-03) (64-bit)`) the demo errors with:

````
Uncaught (in promise) TypeError: GPUDevice.createComputePipeline: 'layout' member of GPUPipelineDescriptorBase is not an object.
````

Passing `None` as layout to `create_compute_pipeline` is perfectly fine and should probably translate to `layout: "auto"` in the actual WebGPU call. Possibly Firefox has not implemented this option yet. This problem should solve itself (nothing we can do now).

File: wonnx/src/gpu.rs [hunk 1]
```diff
@@ -657,46 +661,64 @@ impl GpuTensor {
...
+                buffer_slice.map_async(wgpu::MapMode::Read, |_| {});
+                device.poll(wgpu::Maintain::Wait);
+                buffer_slice.get_mapped_range()
+            };
 
+            let result = Self::read_bytes_to_vec(&output_data[..], shape);
+            drop(output_data);
+            self.buffer.unmap();
+            Ok(result)
+        }
```
haixuanTao, line 699:
```suggestion
            let (tx, rx) = futures_intrusive::channel::shared::oneshot_channel();
            buffer_slice.map_async(wgpu::MapMode::Read, move |v| {
                tx.send(v).unwrap();
            });
            device.poll(wgpu::Maintain::Wait);

            // Awaits until `buffer_future` can be read from
            if let Some(Ok(())) = rx.receive().await {
                let output_data = { buffer_slice.get_mapped_range() };

                let result = Self::read_bytes_to_vec(&output_data[..], shape);
                drop(output_data);
                self.buffer.unmap();
                Ok(result)
            } else {
                Err(GpuError::OutputReadError())
            }
            buffer_slice.map_async(wgpu::MapMode::Read, move |v| {
                tx.send(v).unwrap();
            });
            device.poll(wgpu::Maintain::Wait);

            // Awaits until `MapRead` is effective
            if let Some(Ok(())) = rx.receive().await {
                let output_data = { buffer_slice.get_mapped_range() };

                let result = Self::read_bytes_to_vec(&output_data[..], shape);
                drop(output_data);
                self.buffer.unmap();
                Ok(result)
            } else {
                Err(GpuError::OutputReadError())
            }
```
The `map_async` does not give you any guarantee that its effective without the callback function.

pixelspark: Good catch, although not an issue right now (due to the current implementation) you are technically right and this behaviour could change :+1:

haixuanTao (approved):

Issue #47, opened 2022-02-01
Re-use intermediate buffers


pixelspark:
**Is your feature request related to a problem? Please describe.**

Currently, WONNX will allocate a buffer for each operator output. This output buffer is then read by at least one subsequent operator. After the output has been read by all operators that use it as input, it is not used any longer, but are not deallocated until the 'Session' is dropped (they will be re-used in future inferences). These buffers take up GPU memory, and because GPUs do no swapping as far as I know, they limit the maximum size of a model we can use.

(Note, I am on a MacBook M1 Max with 64 GB memory shared between CPU-GPU so have not run into this issue myself yet)

**Describe the solution you'd like**

Pre-allocating buffers is desirable to ensure inference is fast. This means we should not deallocate buffers after we're done with them (however we'd then also would have to allocate them at inference time). 

As many models are very 'deep', it is very much possible to pre-allocate a smaller number of buffers and re-use these. A simple example graph:

`Input -> A -> B -> C -> Output`

In the above, we currently allocate for Input and outputs of A, B and C. If the output for C fits in the output buffer of A, we could simply reuse A's output buffer for C's output: after B is done reading A's output it will never be used anyway (B must use its own output buffer, as it is still reading from A's output buffer).

A more complicated example:

````
Input -> A 
A -> B -> C
A -> D -> E
C + E -> Output
````

In this case, the output of 'A' is used by both B and D, and can only be re-used after both B and D have executed.

This should be fairly easy to implement by maintaining some sort of 'buffer pool' while sequencing the DAG into GPU operations, and calculating the minimum number and sizes of buffers that should be allocated. This should have some sort of look-ahead to allocate a bigger buffer if an operator further in the graph needs it (so it can be shared with an 'earlier' operator that requires a smaller buffer)

**Describe alternatives you've considered**

That would be one of (1) buying a larger GPU, (2) use smaller models only or (3) implement some sort of swapping...

(I might be able to implement this later on)

haixuanTao:
So, to avoid conflict as you mentioned, I think that we just need to implement a graph coloring algorithm, with the color representing the buffers.

pixelspark:
> So, to avoid conflict as you mentioned, I think that we just need to implement a graph coloring algorithm, with the color representing the buffers.

I am not a CS grad (you might have noticed ;-)) but this seems like the right thing to do. After coloring we then allocate each color's buffer with `max(...buffer sizes requested for this color...)` bytes.

[...]

PR #120, opened 2022-07-03
Fix/readability

pixelspark:
We need to track 'readable' state because wgpu does not allow us to map buffers for reading that are also mappable for writing. We currently do this by passing 'output needs to be readable' flag when traversing the IR DAG.

This PR fixes an issue where if the same node's outputs in the IR DAG are read by more than one node (where it needs to be readable for the 'second' node that reads the output but not for the first), the output buffer is not made readable (as the first node did not require it).

### DIFF
--- a/wonnx/src/gpu.rs
+++ b/wonnx/src/gpu.rs
@@ -109,9 +109,13 @@ impl GpuModel {
         let mut readable_nodes: HashSet<NodeIdentifier> = HashSet::new();
         let mut node_outputs = HashMap::<NodeIdentifier, Vec<GpuTensor>>::new();
         let mut nodes_seen = HashSet::new();
+
+        gpu_model.pre_sequence(root.clone(), &mut readable_nodes, &mut nodes_seen)?;
+        nodes_seen.clear();
+
         gpu_model.sequence(
             root.clone(),
-            &mut readable_nodes,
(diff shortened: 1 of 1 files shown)


pixelspark:
@haixuanTao can you update the CI rust compiler?

````
error: package `wgpu-hal v0.13.1` cannot be built because it requires rustc 1.59 or newer, while the currently active rustc version is 1.58.1
````

(I am on the latest 1.62.0 myself)

haixuanTao:
Feel free to update the CI Rust Compiler.

pixelspark:
OK. I'm not very familiar with the GitHub CI stuff so I will have to dig into this later.

pixelspark:
@haixuanTao the change turns out to be quite trivial, it is part of #119

pixelspark:
Note, this PR basically adds a second step that traverses the DAG (`pre_sequence` followed by `sequence`). There might be ways to fix this issue without the extra pass, but I chose to go this route as it is very likely necessary to have this pass anyway to do the buffer allocation (#47). 

I believe (but perhaps we should make this goal explicit) that we are optimizing for runtime rather than compile time (although compile should of course not be slower than necessary) and graph traversal shouldn't be particularly slow, but anyway. Would like to know your opinion.

PR #122, opened 2022-07-30
docs: add docs.rs badge to readme

philpax:
When looking at a crate's README, I usually look for the docs.rs badge at the top, and assume that it doesn't have docs otherwise. I thought that was initially the case with wonnx, but then I noticed the link to the docs later in the readme. Figured I'd add the badge to stop others from making the same assumption I did 👌

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 <center><img src="logo.svg" alt="WONNX" width="700"/></center>
 
 ![GitHub Workflow Status](https://img.shields.io/github/workflow/status/webonnx/wonnx/CI)
+[![docs.rs](https://img.shields.io/docsrs/wonnx)](https://docs.rs/wonnx)
 ![Crates.io (latest)](https://img.shields.io/crates/dv/wonnx)
 ![Crates.io](https://img.shields.io/crates/l/wonnx)
 


haixuanTao:
Thanks :)

haixuanTao:
Just for info, I'm not sure if the doc on docs.rs is up to date. :S sorry

pixelspark:
It is, but for the last released crate (v0.2.4, 7 months old). We should release a new one!

SludgePhD:
There have been many breaking changes since 0.2.4, so the new version should have been 0.3, not 0.2.5

pixelspark:
> There have been many breaking changes since 0.2.4, so the new version should have been 0.3, not 0.2.5

Not according to semantic versioning: [for major version zero (0.x.y), anything may change at any time, public API is not considered stable](https://semver.org/spec/v0.1.0.html#spec-item-5).

pixelspark:
@SludgePhD v0.2.5 now yanked, v0.3.0 published.

SludgePhD:
Yeah, sadly not doing what Cargo does here would wreak havoc on all downstream crates, so there's not much of a choice

PR #125, opened 2022-08-16
remove incorrect lifetime annotation

aliemjay:
Hi,
rustc was previously accepting this incorrect code:
```rust
async fn test<'a>() { // `'a`, being a lifetime parameter, must be longer than the entire function body
    let f = |_: &'a str| {}; // the argument is required to outlive `'a`
    f(&String::new()); // an argument of shorter lifetime is passed!
}
```
This got fixed in the current beta 1.64, but [we found that wonnx relies on this bug](https://github.com/rust-lang/rust/issues/100544). This PR tries to fix this.

### DIFF
--- a/wonnx/src/ir.rs
+++ b/wonnx/src/ir.rs
@@ -123,7 +123,7 @@ impl<'model> Node<'model> {
         let inputs: Result<Vec<Input<'model>>, IrError> = node
             .get_input()
             .iter()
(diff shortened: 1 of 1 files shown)


haixuanTao:
Thank you ❤️

pixelspark:
Good catch, looks like that was my code 😬

PR #132, opened 2022-11-14
Update to wgpu 0.14

SludgePhD:
Due to https://github.com/gfx-rs/wgpu/issues/3105, multiple versions of wgpu in the same application aren't supported, so update to 0.14 (somehow, using multiple versions sometimes works on my system, but I don't want to unearth any eldritch horrors, so I haven't investigated why).

- In wgpu 0.14, https://github.com/gfx-rs/wgpu/pull/3023 fixes the validation of buffer usages, so they now behave like they do in WebGPU proper. This means we always have to go through an extra (non-storage) buffer when copying data back from the GPU.


haixuanTao:
Nice. Thanks!

PR #139, opened 2023-01-26
operator `LRN` (Local Response Normalization)

CameronTofer:
My attempt at a reference implementation of LRN (Local Response Normalization).


CameronTofer:
I think I've put the .wgsl template in the wrong folder.  @pixelspark, should this template be in endomorphism/ instead of matrix/ ?

pixelspark:
Hm, it doesn't matter that much I think. I don't know what LRN does exactly, so can't decide. Up to you, we can always move it later.

PR #140, opened 2023-02-03
chore: upgrade crates (notably wgpu 0.15)

pixelspark:
Note: WASM version broken because of https://github.com/gfx-rs/wgpu/issues/3430, should wait for new release of wgpu...


pixelspark:
Will make a new one for 0.16 as per #163.

PR #142, opened 2023-02-19
fix: replace 'type' with 'alias' in WGSL

pixelspark:
Fixes the following warning in Chrome:

image

Cannot merge right now, because we are still pinned on wgpu 0.14 (see #140). Applying this now removes the warning in Chrome but makes the native version fail, because wgpu's own wgsl compiler does not yet know about 'alias'.

### DIFF
--- a/wonnx/templates/matrix/gemm.wgsl
+++ b/wonnx/templates/matrix/gemm.wgsl
@@ -1,6 +1,6 @@
-type Scalar = {{ scalar_type }};
-type GemmVec = vec{{ kernel_size }}<{{ scalar_type }}>;
-type GemmMat = mat{{ kernel_size }}x{{ kernel_size }}<{{ scalar_type }}>;
+alias Scalar = {{ scalar_type }};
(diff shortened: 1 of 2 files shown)


FL33TW00D:
Should be fixed when naga! https://github.com/gfx-rs/naga/pull/2299

pixelspark:
Unfortunately https://github.com/gfx-rs/wgpu/issues/3430 has not been fixed yet so we cannot upgrade wgpu just yet (although work has started on fixing the issue this week, from the looks of it)

FL33TW00D:
> Unfortunately [gfx-rs/wgpu#3430](https://github.com/gfx-rs/wgpu/issues/3430) has not been fixed yet so we cannot upgrade wgpu just yet (although work has started on fixing the issue this week, from the looks of it)

From discussions in the matrix channel it is top priority, hopefully a fix will be done this week!

pixelspark:
Superseded by #164

PR #143, opened 2023-02-26
perf: re-use intermediate value buffers

pixelspark:
This PR enabled re-usage of buffers for intermediate values (see #47) using a graph coloring algoritm. Basically, during the 'pre sequence' phase, a buffer manager keeps track of shared buffers and hands them out for use as intermediate value buffer, while keeping track of the lifetimes of each buffer. 

For the included MNIST example (which is basically a linear graph of 8 nodes, optimized down to 6 nodes), we'd previously allocate 6 buffers for intermediate values. With this PR, only 2 are allocated (leading to a reduction of 1.46x in total buffer size). For the Squeeze example, the difference is more dramatic (only 3 buffers needed, 3.19x improvement in size). This is due to the fact that Squeeze too is relatively linear, but with multiple similar blocks of nodes repeated.

Note, this PR should pass all tests but it currently fails with a BERT model:

````
2023-02-26T17:13:46Z ERROR wgpu::backend::direct] Handling wgpu errors as fatal by default
thread 'main' panicked at 'wgpu error: Validation Error

Caused by:
    In a ComputePass
      note: encoder = `<CommandBuffer-(0, 2, Metal)>`
    In a dispatch command, indirect:false
      note: compute pipeline = `bert/embeddings/add`
    Attempted to use buffer with conflicting usages. Current usage STORAGE_READ and new usage STORAGE_READ_WRITE. STORAGE_READ_WRITE is an exclusive usage and cannot be used with any otherusages within the usage scope (renderpass or compute dispatch).
````

Will look into this later.

### DIFF
--- a/wonnx-cli/src/cpu.rs
+++ b/wonnx-cli/src/cpu.rs
@@ -101,7 +101,7 @@ impl Inferer for CPUInferer {
         }
 
         let result = self.model.run(cpu_inputs_ordered)?;
-        log::info!("cpu result: {:?}", result);
+        log::debug!("cpu result: {:?}", result);
 
         let mut output_tensors = HashMap::<String, OutputTensor>::new();
 
@@ -115,7 +115,7 @@ impl Inferer for CPUInferer {
                     .enumerate()
                     .find(|x| &self.model.model.outlet_labels[x.1] == output_name)
                 {
-                    log::info!(
+                    log::debug!(
                         "output node with name '{}' has idx {:?} (and tract id {}, slot {}, name '{}')",
                         output_name,
                         idx.0,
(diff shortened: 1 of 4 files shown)


pixelspark:
OK, so the issue with the larger BERT model is fixed now. On the model I am testing with this change makes a huge difference in memory usage:

````
[2023-02-26T19:12:43Z INFO  wonnx::gpu] shared intermediate value buffers: 5 buffers, total unshared size is 751441920b, total shared size is 13369344b, improvement 56.2063419117647x
````

So about 716 MiB without this change, and 12,75 with 🥇 While this shouldn't really have any impact on inference times, I am seeing decreased inference time as well (from ±19ms to about 14ms per case).

PR #124, opened 2022-07-31
Support basic grouped convolutions

SludgePhD:
Part of https://github.com/webonnx/wonnx/issues/109

Depth-wise convolutions are still failing (eg. https://github.com/onnx/onnx/tree/main/onnx/backend/test/data/pytorch-converted/test_Conv2d_depthwise), I'm not yet sure why. `test_Conv2d_groups` passes with this PR though.

PR #157, opened 2023-04-13
Fix grouped convolutions with 3x3 kernels

SludgePhD:
Followup to https://github.com/webonnx/wonnx/pull/124, fixes the remaining bug.

The bug was caused by the 3x3 convolution optimization triggering for grouped convolutions, which adds padding to the kernel in preparation for `conv_kernel_3x3.wgsl`, but grouped convolutions aren't using that shader, so the default conv shader ended up using the wrong kernel data.

Also includes a fix for the out-of-bounds access check (testing that a `u32` is `< 0` never fails, it has to be done with an `i32`). This wasn't causing any issues as far as I can tell though.


pixelspark:
Great work! :+1: merged

PR #158, opened 2023-04-13
Fix `Clip` node implementation

SludgePhD:
This changes the input->attribute conversion to work with any element type, and passes the min and max attributes to the shader (which was expecting input buffers for some reason).

The upstream tests for this use dynamic min/max inputs and won't work with wonnx, so I've added a custom Rust test.


SludgePhD:
(don't merge yet, I think this breaks `Pad`)

pixelspark:
Nice job :+1: merged

Issue #159, opened 2023-04-13
initializer->attribute promotion doesn't respect `raw_data` field


SludgePhD:
**Describe the bug**
ONNX files can store tensor data either in a type-specific field, or in the `raw_data` field as raw, little-endian bytes.

wonnx promotes constant tensors to node attributes to allow evaluation, but that promotion only looks at the type-specific field, not `raw_data`, leading it to write an empty tensor in the attribute. This then shows up as an error like:

> compiling node 'model_1/model/channel_padding/Pad' failed: the value '[]' is invalid for attribute 'pads' (opset version 13)'

**To Reproduce**
(this requires the optimizer fix from https://github.com/webonnx/wonnx/pull/158 to reproduce)

To reproduce, try running inference for [palm_detection_lite.onnx.zip](https://github.com/webonnx/wonnx/files/11227205/palm_detection_lite.onnx.zip).

PR #160, opened 2023-04-16
Respect `raw_data` when promoting initializer

SludgePhD:
Fixes https://github.com/webonnx/wonnx/issues/159

A better way to avoid issues like this would be to use a proper IR instead of the raw protobuf types, but that's a pretty major refactoring.

I didn't include a test here since the ONNX file in question is pretty large and still doesn't work because of other missing features.

### DIFF
--- a/wonnx/Cargo.toml
+++ b/wonnx/Cargo.toml
@@ -20,7 +20,7 @@ exclude = [
 
 [dependencies]
 wgpu = "0.14.0"
-bytemuck = "1.9.1"
+bytemuck = { version = "1.9.1", features = ["extern_crate_alloc"] }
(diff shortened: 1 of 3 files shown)


pixelspark:
> A better way to avoid issues like this would be to use a proper IR instead of the raw protobuf types, but that's a pretty major refactoring.

@SludgePhD Yes, although then we would lose (or have much more difficulty implementing) copy-on-write (currently IR nodes have a lifetime `'model` just so we can either reference the raw protobufs from the model or newly allocated protobufs using `Cow<'model, SomeProtobufStruct>`). We could gradually refactor `NodeDefinition` to point to fields of the struct instead of the larger struct, but that would not gain us much I think.

PR #162, opened 2023-04-18
Fix upsampling `Resize` op

SludgePhD:


SludgePhD:
@pixelspark Hi :wave: Any chance you could take a look at this PR? It should be fairly straighforward. Thanks!

Issue #169, opened 2023-05-17
Inference without ONNX


philpax:
**Is your feature request related to a problem? Please describe.**
I'm one of the maintainers of the [llm](https://github.com/rustformers/llm) project, and we're looking for a robust, cross-platform GPU inferencing solutions for our LLM models. We currently have computation graphs for GGML, but are planning on installing some kind of abstraction for use with other backends.

I'm investigating the use of `wonnx` as a potential backend, but it is (understandably!) coupled to ONNX. I was wondering if it would be possible to specify a computation graph directly for compilation/inference without going through ONNX.

**Describe the solution you'd like**
A builder API for computation graphs, or something similar, so that a `wonnx::Session` could be created without the use of ONNX.

**Describe alternatives you've considered**
I've considered constructing a `wonnx::onnx::ModelProto` at runtime, but the ONNX format contains a *lot* of things we don't need or don't have. 

It's designed for self-contained models; however, we are loading weights from arbitrary locations and supplying our own computation graph, making it difficult for us to synthesize a complete ONNX model.

**Additional context**
There's no particular hurry on this. We'd love to have GPU inference as soon as possible - especially truly cross-platform, non-CUDA (!) inference - but I assume this would be a large body of work.

I'm also not sure what operations would need to be implemented for our use case, but we would file PRs as required to implement any missing operations.

pixelspark:
Hi @philpax, thanks for bringing this up here. I have given the idea of running LLM’s through wonnx some thought over the past few days and I think it would actually be a great addition (as you say it would provide cross-platform GPU-inference even for non-NVIDIA hardware). 

Adding a builder API would be a good first step. Instead of constructing an ONNX model, it could construct a WONNX IR graph directly (the IR currently is based on nodes that are enums containing mostly ONNX structs, so behind the scenes we would still partially build an ONNX graph, but with a much simpler interface. Eventually we can replace the ONNX structs with our own containing just the bits we need/support).

Ideally LLM makes similar calls to wonnx as it does currently to ggml. 

In short I think we need to implement the following:

1. Builder API to construct WONNX IR directly without (visible) reliance on ONNX.

2. A way to load tensors (initializers in ONNX parlance) from GGML format. This should be easy (at least to implement a slow version) and can be optimized later. 

3. The various ops used in LLM, probably as ‘custom’ non-ONNX standardized operators (i.e. an operator named “ggml.Rope”). 

4. How to handle caches between inference runs.

[...]

PR #170, opened 2023-05-21
Builder API and further removal of ONNX from the IR

pixelspark:
This is a first attempt at addressing #169 as well as an earlier suggestion to move ONNX out of the IR.

tl;dr, this PR exposes a 'builder API' that allows you to use WONNX inference without an ONNX model (of course currently only ops from ONNX are available, but due to the further decoupling of the IR and ONNX, it will be easier to add custom ops in the future):

````rust
use wonnx::builder::*;

let a = tensor("x", &[1, 3], vec![0.1, 0.2, 0.3].into());
let b = tensor("y", &[1, 3], vec![3.0, 2.0, 1.0].into());
let axb = a.add(&b);

let sesh = session_for_outputs(&["result"], &[axb], 13).await.unwrap();
let result = sesh.run(&HashMap::new()).await.unwrap();
assert_eq!(
   result["result"],
   TensorData::F32(vec![3.1, 2.2, 1.3].into())
)
````

Currently the API is very limited (only `add`, `neg` and `conv` ops as well as `input` and `tensor` to define inference-time inputs and weights). Some work that needs to be done:

- [ ] Add all other ops once we are happy with this API 
- [ ] Remove the last bits of ONNX from the IR (`AttributeValue` in particular)
- [ ] Employ the existing shape inference logic for the more complicated ops (currently for `conv` the user has to specify the output shape, but the code to determine the output shape based on input shapes is already in `wonnx_preprocessing`. We might want to move that into the core crate, however the shape inference code is still very specific to ONNX).
- [ ] Discuss overriding operators (so you can do e.g. 'a + b' on two tensors to generate an add.). I did not yet add this now because I think it is better to make things explicit (i.e. in the case of an 'Add' in ONNX you can still set attributes for e.g. broadcasting). 
- [ ] Cleaning up the older graph API and the public API surface in  general
- [ ] Documentation

Help is very much appreciated!

## Nice to haves

- [ ] Expose this API through Python/WASM?


philpax:
This is awesome! Nice work, looking forward to trying it out 🚀

pixelspark:
@philpax you should be able to try this out now (see limitations above). Looking for some feedback!

philpax:
Unfortunately I'm pretty busy this week, but I'll see if one of our devs can try it out 🤞

pixelspark:
> > Our execution happens on the GPU and we generate a single set of commands to execute all ops in one go, so breakpoints are difficult to support. Also intermediate buffers remain in VRAM, are re-used, and cannot easily be made readable.
> > That said, you can mark any node output as inference output (add it to the output list), so you can obtain intermediate results for debugging. This will introduce some copying and extra buffers but is probably useful.
> 
> Alright got it, beeing able to mark any tensor as an output is a nice addition. Would it also be possible to build and execute the commands after each tensor operation and copy the results back from VRAM? It would be highly inefficient but i could see all tensor operation results while debugging. I'm just asking, as i worked with GGML a bit and debugging it is an ..... _experience_.

Well, given the current API, you could build your graph iteratively, create a new session each time (with the most recently added node as output) and call `.run` on it. (The `TensorRef` you get from the graph builder API are not tied to a session and so can be re-used. A session does exclusively maintain GPU buffers though). It would be terribly slow (also because intermediate results will not be re-used, so your are calculating 1 + 2 + 3 + 4 + .. + N ops instead of just N) but I've used this approach in the past with some succes (let it generate a file containing each nodes output, then use a few scripts to find differences between a known reference, stop execution when the difference is too large, etc.). 

Another way to do it would be to generate a graph and session for a single op each time (and then run that with the previous node's output). This is a bit more performant but you are still copying everything back and forth from the GPU of course. Also you need to somehow deal with ops that take more than one input (but given the relative simplicity of the LLaMA models that should be doable).

LLukas22:
Hey i finally got some time to take a look at these changes and i will start to play around a bit. Is there somewhere a list of all supported Tensor ops? And how would i integrate a custom OP, like Alibi?

pixelspark:
> Hey i finally got some time to take a look at these changes and i will start to play around a bit. Is there somewhere a list of all supported Tensor ops? And how would i integrate a custom OP, like Alibi?

Currently it is probably the easiest to check the giant `match` statement in `wonnx/src/compiler.rs`. It matches by ONNX op name and then generates shader code (through tera templates). I want to change this to an enum/trait and associated methods in this PR but haven't yet had the time to do so. 

To implement a new op, see [these instructions](https://github.com/webonnx/wonnx#contribution-on-implementing-a-new-operator). Remember that in addition to implementing the op itself we also prefer to have some tests and (if relevant) shape inference code.

LLukas22:
I converted a GPT-2 model to ONNX and visualized the operations [here](https://github.com/webonnx/wonnx/assets/65088241/676c145f-5c21-48d3-ab19-aec523fb628c). The main thing still missing are the different shape manipulating operations like "Squeeze, Unsqueez, Concat, Gather" and stuff like that. But maybe you could also take a look, this is the first time im working with onnx, so maybe im missing something😅.

I will probably try to implement a multihead-attention layer with the current graph api.

pixelspark:
> I converted a GPT-2 model to ONNX and visualized the operations [here](https://github.com/webonnx/wonnx/assets/65088241/676c145f-5c21-48d3-ab19-aec523fb628c). The main thing still missing are the different shape manipulating operations like "Squeeze, Unsqueez, Concat, Gather" and stuff like that. But maybe you could also take a look, this is the first time im working with onnx, so maybe im missing something😅.

Forgot to mention two things that might be useful:

* You can use the `nnx` (`wonnx-cli`) tool to list operations used in an ONNX model (it is not smart enough to check WONNX support itself). That is a bit easier to read than your chart I guess :-)
* Some operations (`Reshape`, `Squeeze`, `Unsqueeze` notably) only modify shape information and are 'optimized away' in `wonnx/src/optimizer.rs` (removed from the DAG).

From the top of my head Concat and Gather are implemented (but if not, relatively trivial operations as they just copy data)

LLukas22:
> Forgot to mention two things that might be useful:
> 
>     * You can use the `nnx` (`wonnx-cli`) tool to list operations used in an ONNX model (it is not smart enough to check WONNX support itself). That is a bit easier to read than your chart I guess :-)
> 
>     * Some operations (`Reshape`, `Squeeze`, `Unsqueeze` notably) only modify shape information and are 'optimized away' in `wonnx/src/optimizer.rs` (removed from the DAG).

Yeah i will give this a try, maybe it will produce a more readable output. But GPT-2 should be a good starting point as it is relatively simple, openly available and contains all building blocks we need for other LLMs.  


> From the top of my head Concat and Gather are implemented (but if not, relatively trivial operations as they just copy data)

I just looked at the functions implemented by `TensorRef`, which isn't much but later realized  these are just wrappers around the functions defined in `compiler.rs`. But i now know what you were referring to with "Replace "op name + attributes" with an actual enum containing all (supported) ops and their parameters". 😅 


I'm also kinda confused how i could create a `f16` tensor,  as the `TensorData` enum only specifies `F32`. Would the existing operations even work with `f16` tensors?

LLukas22:
Got it, my plan was to somehow get a GPT-2 F16 GGML model loaded and executed and then work my way toward some more exotic quantization formats, which will require more work. 

Is there a way to instantiate a tensor from a memory address, an offset and dimensions? Or do i have to first load the ggml tensor into lists and then instantiate a tensor with these lists? Would something like direct GPU storage be possible or is this out of scope of the web gpu standard?

FL33TW00D:
> 

If I were you, I would write a script that packs the GGML weights. This would pack 2 Float 16 values into a single u32 value. You could then use https://www.w3.org/TR/WGSL/#unpack2x16float-builtin to unpack them inside your shader.

LLukas22:
>If I were you, I would write a script that packs the GGML weights. This would pack 2 Float 16 values into a single u32 value. You could then use https://www.w3.org/TR/WGSL/#unpack2x16float-builtin to unpack them inside your shader. 

This could be interesting, thanks. But i thought WebGPU had native `f16` support since [this](https://github.com/gpuweb/gpuweb/pull/2696) got merged. Thats the reason i was asking for `f16` support in WONNX. 

> I've done something similar here: https://gist.github.com/FL33TW00D/d81562557279d887705985f7c6ae4481
But it operates on an existing ONNX model.

Also very interesting, i probably have to try something similar when i try to implement the other quantization formats. But i guess i can also reference the BLAS GGML implementation a bit.

FL33TW00D:
> > If I were you, I would write a script that packs the GGML weights. This would pack 2 Float 16 values into a single u32 value. You could then use https://www.w3.org/TR/WGSL/#unpack2x16float-builtin to unpack them inside your shader.
> 
> This could be interesting, thanks. But i thought WebGPU had native `f16` support since [this](https://github.com/gpuweb/gpuweb/pull/2696) got merged. Thats the reason i was asking for `f16` support in WONNX.
> 
> > I've done something similar here: https://gist.github.com/FL33TW00D/d81562557279d887705985f7c6ae4481
> > But it operates on an existing ONNX model.
> 
> Also very interesting, i probably have to try something similar when i try to implement the other quantization formats. But i guess i can also reference the BLAS GGML implementation a bit.

`gpuweb/gpuweb` hosts the specification, but that does not mean it has been implemented anywhere.
F16 is gated behind a feature gate, and this is not currently available in any stable browsers. 

Chrome is in the process of shipping F16, but not yet. `naga` and `wgpu` support will take even longer, and this would be required for F16 in WONNX.

pixelspark:
> Is there a way to instantiate a tensor from a memory address, an offset and dimensions? Or do i have to first load the ggml tensor into lists and then instantiate a tensor with these lists? 

The `TensorData` enum takes a `Cow<'model, [f32]>` so it will happily use any slice you can throw at it as long as it has a lifetime of `'model` or better (use `bytemuck` as used elsewhere in wonnx to cast slices if necessary. I suppose you will have a pointer to some mmap'ed memory you want to convert into a slice). 

> Would something like direct GPU storage be possible or is this out of scope of the web gpu standard?

Not sure what you mean by this - as far as I know there are no special provisions for unified memory in WebGPU/wgpu but I might be mistaken.

LLukas22:
> Not sure what you mean by this - as far as I know there are no special provisions for unified memory in WebGPU/wgpu but I might be mistaken.

I was refering to something like [GPUDirect Storage](https://developer.nvidia.com/gpudirect-storage) where data can be moved directly into VRAM without beeing loaded into RAM first.

>The TensorData enum takes a Cow<'model, [f32]> so it will happily use any slice you can throw at it as long as it has a lifetime of 'model or better (use bytemuck as used elsewhere in wonnx to cast slices if necessary. I suppose you will have a pointer to some mmap'ed memory you want to convert into a slice).

If i find some time this weekend i'll give it a try. Thanks for the info. And yes you are right we are working with either mmap'ed memory or a pointer to a file. 

As i don't want to spam this thread further, is there a way to reach you directly (e.g. discord) if i have further questions? And again sorry for beeing nooby, but as i said this is my first time working with wgpu.😅

pixelspark:
> As i don't want to spam this thread further, is there a way to reach you directly (e.g. discord) if i have further questions? And again sorry for beeing nooby, but as i said this is my first time working with wgpu.😅

Feel free to spam the thread ;-) I am not (regularly) on Discord but can be reached by email (see profile).

PR #175, opened 2023-07-18
Add argmax operator.

JulienSiems:
This PR adds the argmax operator to wonnx.

### DIFF
--- a/wonnx/tests/reduce.rs
+++ b/wonnx/tests/reduce.rs
@@ -62,15 +62,16 @@ fn reduce() {
         60.0, 2.0,
     ];
 
-    let data2 = [
-        5.0, 1.0, 
-        2.0, 20.0, 
+    #[rustfmt::skip]
+    let data_two = [
+        20.0, 1.0, 
+        5.0, 2.0, 
         
         30.0, 1.0, 
         40.0, 2.0, 
         
-        1.0, 55.0,
-        60.0, 2.0,
+        60.0, 1.0,
+        55.0, 2.0,
     ];
 
     // ReduceSum: sum all
@@ -236,12 +237,12 @@ fn reduce() {
 
     // ONNX test case: do_not_keepdims with ArgMax
     test_reduce(
-        &data2,
+        &data_two,
         &[3, 2, 2],
         Some(vec![1]),
         "ArgMax",
         false,
-        &[1., 2., 1., 1., 2., 1.],
+        &[0., 1., 1., 1., 0., 1.],
         &[3, 2],
     );
 


wonnx/templates/pool/reduce.wgsl
```diff
@@ -40,7 +40,9 @@ fn main(@builtin(global_invocation_id) global_id: vec3<u32>) {
 		Now for each reduced axis, iterate all values and reduce. Note, starting value may not always be zero. For 
 		ReduceMin/Max we should initialize as NaN and keep a flag to check if we have seen at least one element -#}
 
-		var accumulator = {% if op_type == "ReduceProd" %} {{ scalar_type }}(1) {% else %} Scalar() {% endif %}; 
+		var accumulator = {% if op_type == "ReduceProd" %} {{ scalar_type }}(1) {% else %} Scalar() {% endif %};
+	    var max_element: Scalar = log(Scalar());
```
pixelspark, line 44: Hm, perhaps add a note explaining why you initialize to `log(Scalar())` (I assume it is a trick to initialize to `-Inf` so the first value is always higher).

wonnx/templates/pool/reduce.wgsl
```diff
@@ -79,6 +81,11 @@ fn main(@builtin(global_invocation_id) global_id: vec3<u32>) {
 					else if(accumulator < input_val) {
 						accumulator = input_val;
 					}
+				{% elif op_type == "ArgMax" %}
+					if(input_val > max_element) {
+                        max_element = input_val;
+                        accumulator = f32(count);
```
pixelspark, line 87: The output of `ArgMax` is actually specified as `int64` (see [here](https://onnx.ai/onnx/operators/onnx__ArgMax.html)). This should work for now but we might have to fix it later.

wonnx/tests/reduce.rs
```diff
@@ -223,6 +235,17 @@ fn reduce() {
         &[3, 2],
     );
 
+    // ONNX test case: do_not_keepdims with ArgMax
```
pixelspark, line 238: Can you also enable the corresponding test case in the ONNX backend test (Python scripts)?

pixelspark (changes requested):
Great contribution! Just a few notes, should be fine to merge after.

wonnx/templates/pool/reduce.wgsl
JulienSiems: Good point, will do!

wonnx/templates/pool/reduce.wgsl
JulienSiems: Good point, then it would probably make sense to move argmax out of reduce to not have to return mixed data types from reduce?

wonnx/tests/reduce.rs
JulienSiems: Will do!

PR #180, opened 2023-07-30
Allow larger output sizes for Concat

### LINKED ISSUE #178: ComputeLimitExceeded("X threads", 96800, 65535)
**Describe the bug**
I get the error `ComputeLimitExceeded("X threads", 96800, 65535)` when loading a network that concatenates two tensors.
The inputs are two tensors with shape [1, 64, 440, 440].

**To Reproduce**
Load a network that has a similar input size.
My ONNX sample input wouldn't be helpful here, because it uses a ConvTranspose operation which I am working on adding to wonnx in my fork.

**Expected behavior**
The concatenation should work.

**Screenshots**
I added some output to print the input and output shapes in the compiler:
```
[2023-07-23T19:54:42Z DEBUG wonnx::compiler] [
        Shape {
            dims: [
                1,
                64,
                440,
                440,
            ],
            data_type: F32,
        },
        Shape {
            dims: [
                1,
                64,
                440,
                440,
            ],
            data_type: F32,
        },
    ]
[2023-07-23T19:54:42Z DEBUG wonnx::compiler] [
        Shape {
            dims: [
                1,
                128,
                440,
                440,
            ],
            data_type: F32,
        },
    ]
```

Is this a general limitation and an indicator that the network might be too complex for wonnx or would it make sense to consider bumping the workgroup size or maybe making it dynamic depending on the input size?

mayjs:
UNet caused errors because the Concat operator could not handle the large output sizes of some Concat operations.
The solution for this is to change the workgroup dimensions to also use the y dimension.
Since there is a limit of threads per dimension, this allows us quadratically increase the maximum possible output size.

Fixes #178.

### DIFF
--- a/wonnx/src/compiler.rs
+++ b/wonnx/src/compiler.rs
@@ -2,7 +2,7 @@
 use crate::utils::{
     ceil, AttributeNotFoundError, DataTypeError, MultiType, NodeAttributes, ScalarType, Shape,
 };
-use num::integer::gcd;
+use num::integer::{gcd, Roots};
 use tera::{Context, Tera};
 use thiserror::Error;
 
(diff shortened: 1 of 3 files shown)


pixelspark:
:+1: Let's await the CI run but otherwise good to merge

PR #183, opened 2023-08-05
Fix Concat for larger inputs and add an additional test

mayjs:
I messed this up in !180, sorry about that.
The `num_workgroups` obviously only contains the number of workgroup invocations along each axis.
To calculate `actual_idx`, we need to consider the total number of executions in the x axis, so we have to multiply with the workgroup size.

### DIFF
--- a/wonnx/templates/matrix/concat.wgsl
+++ b/wonnx/templates/matrix/concat.wgsl
@@ -17,9 +17,9 @@ fn main(@builtin(global_invocation_id) global_id: vec3<u32>, @builtin(num_workgr
 	let gidx = global_id.x;
         let gidy = global_id.y;
 
-        let nx = num_workgroups.x;
+        let x_executions = num_workgroups.x * 16u;
 
-        let actual_idx = gidx + gidy * nx;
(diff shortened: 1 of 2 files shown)


mayjs:
Just as a side note: With this fix and the current state of !182 I can successfully execute a UNet network in wonnx.

pixelspark:
Cool! It seems our CI tests are not configured properly but the code looks good. Would be nice to have UNet as a (testable) example, is that something you could add easily?

mayjs:
> Cool! It seems our CI tests are not configured properly but the code looks good. Would be nice to have UNet as a (testable) example, is that something you could add easily?

Well, that depends on how functional it would have to be for the example.
The pretrained model I'm using right now is GPL licensed, so we can't integrate it into this repository.
Besides, the ONNX file is about 60MB, so it would not really be a great idea to integrate it into the repository as is.

Maybe we could somehow generate the UNet graph in code and initialize the weights with a random number generator and a fixed seed?
That way we could at least test for the correct output values, but the network wouldn't do anything useful of course.

PR #199, opened 2023-12-21
Erf operator  Implementation

AmineDiro:
Hello,

This is implementation on the approximate [Erf operator](http://www.xavierdupre.fr/app/mlprodict/helpsphinx/onnxops/onnx__Erf.html#erf-13). This operator is needed for models using the [GELU](https://pytorch.org/docs/stable/generated/torch.nn.GELU.html) activation function for example the majority of transformer models.

Onnx supports Gelu directly but it is starting from opset 20 so I think that this operator is useful for now.

Happy to make changes based on your suggestions !


pixelspark:
Hi @AmineDiro! Sorry, only just got around to having a look at this. Your PR looks good! I would suggest however placing the Erf calculation itself in the activation templates ([here](https://github.com/webonnx/wonnx/tree/master/wonnx/templates/snippets)) - otherwise we will end up with many different templates for similar functions.

AmineDiro:
Hi @pixelspark , No worries, thanks for taking the time to review this PR!
I didn't know where to put the function I'll do the necessary modification and push it! Thanks again and happy new year 🎉

sluramod:
@pixelspark could you please merge this PR?

pixelspark:
Sorry, took a while, thanks for the contribution!

PR #202, opened 2024-03-05
Improve `DataTypeError` message with `parametrized dimension`

Abioy:
mention the field name in error message as well as the index of the parametrized dimension, so user can fix it


pixelspark:
Thanks!

PR #204, opened 2024-03-19
set async_recursion to ?Send for wasm32 arch

astnmsn:
In the dependency upgrade for wgpu from 16 -> 19, wbgpu started properly labeling their structs as non send/sync which broke the expectations for the async_recursion proc macro. By configuring the macro to not expect a type that is Send, the wasm build of wgpu should properly compile. No callsites of the updaded function require that the type be Send, so there wasn't a need to update anything else.


pixelspark:
Thanks for sorting this out @astnmsn 💯

PR #214, opened 2024-07-21
Implement the slice operator and fix the cancat operator

kadu-v:
This is my first contribution to wonnx.
I implemented the Slice operator and fixed the Concat operator. Additionally, I implemented the yolox_nano sample. 
The Slice operator only supports cases where there is a single axis.
Please let me know if there are any points that need to be corrected before merging.

On a side note, I was able to run a YOLOX app on an iPhone using wonnx. [Here](https://github.com/kadu-v/wonnx-ios) is the repository.


pixelspark:
@kadu-v very cool to see WONNX in a native iOS app! (I'm sure Apple by now has APIs that provide functionality similar to YOLOX and use more efficient hardware, i.e. the ANE, but still this might be useful as it offers more control over the model used). 

Will have a look at this PR later!

